Issue #25, opened 2022-10-27
Spec doesn't mention what happens if child trie key is requested


tomaka:
The spec doesn't mention what happens if [chainHead_unstable_storage](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_unstable_storage.html#chainhead_unstable_storage) is called, and that `key` corresponds to a child trie, and that `child_key` is null.

tomaka:
Or vice-versa: if `child_key` is non-null and `key` isn't a child trie.

tomaka:
I've learned this morning that it was "forbidden" for runtimes to write any key that starts with `:child_storage:`
("forbidden" is in quotes because the write is actually ignored)

Consequently, there's no ambiguity about whether something is a child storage or not.

If the JSON-RPC client wants to read a child storage key, it can simply read `concat(":child_storage:default:", main_trie_key, child_key)`.

We can thus remove the `child_key` parameter altogether.

tomaka:
I think the most sensible option is to do it the same way as the `storage_get` host function, which is also the same way as the networking protocol.
In other words, treat as if the child trie entry in the main trie was a regular entry.

PR #26, opened 2022-11-09
Clarify behavior of `chainHead_unstable_storage` w.r.t. child tries

tomaka:
Close https://github.com/paritytech/json-rpc-interface-spec/issues/25
cc @cheme 

I went with the decision that querying a trie entry that starts with `:child_storage:` must always return `null`, to avoid confusions.


cheme:
Could be fine. Then we need a specific rpc to get child trie root, guess it is better (no need to know the prefix).

Thinking out loud, in a similar way, all rpc for child trie are using prefixed storage key to select child trie and require knowledge of this prefixing, maybe the rpc should work the same way as the host function and be specialized for the default child trie kind (then storage_key will be the non prefix child trie location which would be more user friendly).
Actually rpc for child trie should be also able to query from a given child trie root, but this also require to know the kind of the child trie, so also a rpc specialized for default child trie only.

cheme:
On a second thought, the rpc is querying the state and the state does contain a value, so not sure what the most confusing behavior is.

cheme:
well I don't understand the rational here

src/api/chainHead_unstable_storage.md
```diff
@@ -5,12 +5,16 @@
...
 
 **Return value**: String containing an opaque value representing the operation.
 
-The JSON-RPC server must start obtaining the value of the entry with the given `key` (and possibly `childKey`) from the storage.
+The JSON-RPC server must start obtaining the value of the entry with the given `key` from the main trie of the storage. If `childTrie` is not `null`, then the entry must instead be loaded from the child trie whose key is `childTrie`. If the requested child trie doesn't exist, it should be treated as if it was an empty trie.
+
+> **Note**: When `childTrie` is `null`, querying a `key` that starts with the ASCII-encoded string `:child_storage:` *must* produce a `null` value. This behavior is different from the behavior of the `ext_storage_get_version_1` host function in the client-host interface, where querying a key starts with the ASCII-encoded string `:child_storage:` might return the hash of the root of a child trie.
+
+> **Note**: The `childTrie`, if not `null`, doesn't need to be prefixed with `:child_storage:` or `:child_storage:default:`. In other words, any child trie key is potentially valid, and not just ones that start with a specific prefix.
```
cheme, line 17:
```suggestion
> **Note**: The `childTrie`, if not `null`, doesn't need to be prefixed with `:child_storage:default:`. In other words, any child trie key is a valid default child trie key, and not just ones that start with a specific prefix.
```
or something similar, I mean the tradeoff with not having the prefix is that we target only `:child_storage:default:` and nothing else.

tomaka:
This API is really designed to be used by UIs. In order to know what a certain storage value is, the UI looks in the metadata (or, alternatively, already knows what the key is) then uses this function.

Consequently:

- I don't think there's a need to support anything else but default child tries. If non-default child tries are added, we can always bump the API to v2.
- I don't think there's any need to know the root of a child trie. Like, what would be the point of knowing that for a JSON-RPC client?

cheme:
>  I don't think there's a need to support anything else but default child tries. If non-default child tries are added, we can always bump the API to v2.

Agree totally (and non-default child state/trie would be likely to have a slightly different api depending on their nature).

> I don't think there's any need to know the root of a child trie. Like, what would be the point of knowing that for a JSON-RPC client?

JSON-RPC client could be local user of a chain. I mean this is the main point of access for external tools, not just UI. For instance there are debug tools in polkadot-js that are pretty useful.

But out of this question, I don't see the point of hiding some state values.
(I mean we are not filtering all values that are not define in metadata)

jsdw (approved):
I resolved the conflicts and FWIW I think this clarification is quite handy so happy to see it merge if you guys are ok with it!

lexnv (approved):

tomaka:
This was meant to be superseded by https://github.com/paritytech/json-rpc-interface-spec/pull/67

Issue #27, opened 2022-11-24
Notes about the design of archive_ functions


tomaka:
This issue consists in some random notes regarding the `archive_` functions:

- Contrary to `chainHead_`, I don't think we need to make them cancellable. Implementations of the `archive_` functions would typically consist either in looking into some database, or querying some trusted remote node. In both cases, and contrary to the `chainHead_` functions, obtaining the information consists in a single request, and so there isn't really much benefit in cancelling an ongoing request.

- The API should make explicit which blocks can be queried and which can't. I suppose that a simple `archive_finalizedBlock` function that returns just the finalized block should do the trick. Only blocks that are ancestors of the finalized block could be queried. There's no need for subscriptions, as you're not supposed to use `archive_` to query things at the head of the chain.

- The design of this API should not forget the fact that we may want to be able to query blocks that the client can't verify anymore, as explained in https://forum.parity.io/t/deprecating-older-polkadot-kusama-blocks/1447. This means that it should be implementable for example on a proxy.

PR #36, opened 2023-05-15
Design the `archive` API

tomaka:
Close https://github.com/paritytech/json-rpc-interface-spec/issues/27

This PR designs the `archive` API.

The API now exposes the concept of "finalized block". If you query any block inferior or equal to that finalized block, then you're guaranteed to get one result (and only one).

If you query any block superior to that finalized block, then you might get multiple results and blocks might disappear and everything is wacko.
The `archive` functions aren't meant to follow the head of the chain, but I think that it is not a bad idea to allow querying non-finalized blocks especially in the case where finality is stalled for some reason.
It is possible and allowed for an implementation to decide to return only finalized blocks, in other words querying a block superior to the finalized block always returns `null` (or an empty array for `archive_unstable_hashByHeight`). This is important for load balancing and caching purposes.

Note that `archive_unstable_finalizedHeight` isn't a subscription, because you're not supposed to call it periodically. It's a function the client is expected to call once at initialization or once every 5 minutes. Again, this is for archiving, not for following the head of the chain.

I have removed the `networkConfig` and the subscription system in the name of keeping things simple and not over-engineering. While for `chainHead` this is important, I think that these functions will never be implemented by a light client. These functions are supposed to be slow and called infrequently on very heavy full nodes, rather than something swift. As such, being able to cancel an operation is in my opinion not important.


tomaka:
I have now realized that `includeDescendants` of `archive_unstable_storage` needs a mechanism similar to `state_getKeysPaged`.

If https://github.com/paritytech/json-rpc-interface-spec/pull/37 is accepted, I will turn `archive_unstable_storage` into a subscription and add the same mechanism where the client has to call a `continue` function after a certain number of items.

src/api/archive_unstable_body.md
```diff
@@ -1,3 +1,11 @@
 # archive_unstable_body
 
-**TODO**
+**Parameters**:
+
+- `hash`: String containing the hexadecimal-encoded hash of the body to retreive.
```
skunert, line 5:
```suggestion
- `hash`: String containing the hexadecimal-encoded hash of the body to retrieve.
```
Nit: Its pretty clear what is meant here, but for first-time readers it may sound like they should  supply the hash of the body instead of the hash of the header here.

skunert (commented):
Looks good, not using subscriptions for these sounds reasonable to me.

skunert (approved):

lexnv (approved):

skunert (approved):
Looks good, @lexnv anything from your side (since there were changes since last review)? Otherwise we can merge.

skunert (approved):
Looks good, @lexnv anything from your side (since there were changes since last review)? Otherwise we can merge.

Issue #21, opened 2022-09-14
Is it planned to support storage iterations?


bkchr:
Is it planned to support any kind of storage iterations? Similar to the current `state_getKeysPaged` in Substrate?

tomaka:
It's not planned but we can add it.

bkchr:
Yeah, at least having it in `archive_` would be nice. Would you see this/like to see that we have here all the things documented in this spec that are related to JSON rpc in Substrate? Probably yes or? 
There is maybe some other stuff that needs to be documented here, otherwise people will be unhappy with the new api probably. 

> I haven't implemented "getting the next key" anywhere in the light client yet, so I'm not sure how much networking it involves.

That is a good question. I think the main question would here be, can we build UIs without this?

tomaka:
> Would you see this/like to see that we have here all the things documented in this spec that are related to JSON rpc in Substrate? Probably yes or?

Sorry I don't get the question.

> I think the main question would here be, can we build UIs without this?

To me, the JSON-RPC functions that allow you to access the storage are here to allow you to access the keys indicated by the metadata. For example, if the metadata tells you "information X is at key K", you then try to read key K.
As far as I know, the metadata tells you that there are maps, and so it makes sense to give the possibility to list all the elements of a map, in other words all the keys that start with a certain prefix.

So I don't think that "get the next key" makes sense, however "get all the keys that start with a certain prefix" makes sense to me.
I've already implemented this in the light client by asking for a Merkle proof of the prefix and then following the items in the proof.

bkchr:
> Sorry I don't get the question.

What I meant is that there are maybe some other RPC functions which are not yet specified here. So, that we may also have some other functions that need to be described here.

> So I don't think that "get the next key" makes sense, however "get all the keys that start with a certain prefix" makes sense to me.
> I've already implemented this in the light client by asking for a Merkle proof of the prefix and then following the items in the proof.

In the end probably boils down to the same. I think one of the reasons for having this kind of "iteration" support on the JSON rpc level is that there are maps that are too big to send at once, so you do multiple requests to fetch them. The best would be some kind of stream to poll for elements (you could still fetch all keys at once with a storage proof for example). However, this is probably not supported by JSON rpc?

PR #37, opened 2023-05-15
Make `chainHead_unstable_storage` more powerful

tomaka:
Close https://github.com/paritytech/json-rpc-interface-spec/issues/21

Even though they're unrelated, this PR is based on top of https://github.com/paritytech/json-rpc-interface-spec/pull/35. I expect https://github.com/paritytech/json-rpc-interface-spec/pull/35 to be merged, and didn't want to have conflicts.

This PR enhances `chainHead_unstable_storage` by adding back the `type` field that was removed in #17.
You can now query not just the value of a storage item, but also the hash of its value, the values of the item and all its descendants, the hashes of item and all its descendants, or the Merkle value of the item or its closest ancestor.

Querying the descendants of the storage item makes it possible to enumerate all the keys under a certain prefix, which is important to know the content of maps.
In order to reproduce something similar to `state_getKeysPaged`, in the case when you enumerate descendants the server can now generate a `waiting-for-continue` event, after which the client must call `chainHead_unstable_storageContinue` in order to resume the enumeration. This makes it possible for the client to interrupt the enumeration at any point in time. Rather than letting the client decide of the number of items (like is the case for `state_getKeysPaged`), it's now the server which decides, which makes it possible to decide to generate this event based on the size of what was sent rather than the number of items.

The `closest-ancestor-merkle-value` query type makes it possible to know when the content of a map has changed. It can be seen as similar to `hash`, except that `hash` is only the hash of the value while `closest-ancestor-merkle-value` is the hash of the value plus the hash of all the descendants in the trie.
Basically, in order to watch a map, a JSON-RPC client would periodically query the `closest-ancestor-merkle-value` of the map, then redownload the list of keys if this Merkle value changes.

### DIFF
--- a/src/api/chainHead_unstable_storage.md
+++ b/src/api/chainHead_unstable_storage.md
@@ -5,12 +5,12 @@
 - `followSubscription`: An opaque string that was returned by `chainHead_unstable_follow`.
 - `hash`: String containing an hexadecimal-encoded hash of the header of the block whose storage to fetch.
 - `key`: String containing the hexadecimal-encoded key to fetch in the storage.
-- `childKey`: `null` for main storage look-ups, or a string containing the hexadecimal-encoded key of the trie key of the trie that `key` refers to. **TODO**: I don't know enough about child tries to design this properly
+- `childTrie`: `null` for main storage look-ups, or a string containing the hexadecimal-encoded key of the child trie of the "default" namespace.
 - `networkConfig` (optional): Object containing the configuration of the networking part of the function. See [here](./api.md) for details. Ignored if the JSON-RPC server doesn't need to perform a network request. Sensible defaults are used if not provided.
 
 **Return value**: String containing an opaque value representing the operation.
 
-The JSON-RPC server must start obtaining the value of the entry with the given `key` (and possibly `childKey`) from the storage.
+The JSON-RPC server must start obtaining the value of the entry with the given `key` from the storage, either from the main trie of from `childTrie`.
 
 The operation will continue even if the given block is unpinned while it is in progress.
 
--- a/src/api/chainHead_unstable_storage.md
+++ b/src/api/chainHead_unstable_storage.md
@@ -67,7 +67,7 @@ If the `type` parameter was `"value"` or `"descendants-values"`, then the `value
 
 If the `type` parameter was `"hash"` or `"descendants-hashes"`, then the `hash` field is set. The `hash` field a string containing the hexadecimal-encoded hash of the storage entry.
 
-If the `type` parameter was `"closest-ancestor-merkle-value"`, then the `merkle-value` field is set and the `key` field indicates which closest ancestor has been found. The `merkle-value` field a string containing the hexadecimal-encoded Merkle value of the storage entry or its closest ancestor.
+If the `type` parameter was `"closest-ancestor-merkle-value"`, then the `merkle-value` field is set and the `key` field indicates which closest ancestor has been found. The `merkle-value` field a string containing the hexadecimal-encoded Merkle value of the storage item indicated by the `key` field.
 
 Only one of `value`, `hash` or `merkle-value` are set at any given time.
 
--- a/src/api/chainHead_unstable_storage.md
+++ b/src/api/chainHead_unstable_storage.md
@@ -11,7 +11,7 @@
 
 **Return value**: String containing an opaque value representing the operation.
 
-The JSON-RPC server must start obtaining the value of the entry with the given `key` from the storage, either from the main trie of from `childTrie`. If `type` is `descendants-value` or `descendants-hashes`, then it must also obtain the values of all the descendants of the entry.
+The JSON-RPC server must start obtaining the value of the entry with the given `key` from the storage, either from the main trie or from `childTrie`. If `type` is `descendants-value` or `descendants-hashes`, then it must also obtain the values of all the descendants of the entry.
 
 The operation will continue even if the given block is unpinned while it is in progress.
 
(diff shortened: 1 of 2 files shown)


src/api/chainHead_unstable_storage.md
```diff
@@ -5,19 +5,28 @@
...
+- `childTrie`: `null` for main storage look-ups, or a string containing the hexadecimal-encoded key of the child trie of the "default" namespace.
+- `type`: String equal to one of: `value`, `hash`, `closest-ancestor-merkle-value`, `descendants-values`, `descendants-hashes`.
 - `networkConfig` (optional): Object containing the configuration of the networking part of the function. See [here](./api.md) for details. Ignored if the JSON-RPC server doesn't need to perform a network request. Sensible defaults are used if not provided.
 
 **Return value**: String containing an opaque value representing the operation.
 
-The JSON-RPC server must start obtaining the value of the entry with the given `key` (and possibly `childKey`) from the storage.
+The JSON-RPC server must start obtaining the value of the entry with the given `key` from the storage, either from the main trie of from `childTrie`. If `type` is `descendants-value` or `descendants-hashes`, then it must also obtain the values of all the descendants of the entry.
```
lexnv, line 14:
```suggestion
The JSON-RPC server must start obtaining the value of the entry with the given `key` from the storage, either from the main trie or from `childTrie`. If `type` is `descendants-value` or `descendants-hashes`, then it must also obtain the values of all the descendants of the entry.
```

src/api/chainHead_unstable_storage.md
```diff
@@ -35,18 +44,56 @@ This function will later generate notifications in the following format:
...
+If the `type` parameter was `"value"`, `"hash"`, `"descendants-values"` or `"descendants-hashes"`, this `key` is guaranteed to start with the `key` provided as parameter.
+If the `type` parameter was `"value"` or `"hash"`, then it is also guaranteed to be equal to the `key` provided as parameter.
+If the `type` parameter was `"closest-ancestor-merkle-value"`, then the `key` provided as parameter is guaranteed to start with the value in the `key` field.
+
+If the `type` parameter was `"value"` or `"descendants-values"`, then the `value` field is set. The `value` field a string containing the hexadecimal-encoded value of the storage entry.
+
+If the `type` parameter was `"hash"` or `"descendants-hashes"`, then the `hash` field is set. The `hash` field a string containing the hexadecimal-encoded hash of the storage entry.
+
+If the `type` parameter was `"closest-ancestor-merkle-value"`, then the `merkle-value` field is set and the `key` field indicates which closest ancestor has been found. The `merkle-value` field a string containing the hexadecimal-encoded Merkle value of the storage item indicated by the `key` field.
```
lexnv, line 70: small nit:
```suggestion
If the `type` parameter was `"value"` or `"descendants-values"`, then the `value` field is set. The `value` field is a string containing the hexadecimal-encoded value of the storage entry.

If the `type` parameter was `"hash"` or `"descendants-hashes"`, then the `hash` field is set. The `hash` field is a string containing the hexadecimal-encoded hash of the storage entry.

If the `type` parameter was `"closest-ancestor-merkle-value"`, then the `merkle-value` field is set and the `key` field indicates which closest ancestor has been found. The `merkle-value` field is a string containing the hexadecimal-encoded Merkle value of the storage item indicated by the `key` field.
```

lexnv (approved):
Amazing! 

This should give us parity between what substrate exposed previously wrt iterating over key storages, while having a more compact and robust API.

Hopefully, we'll be able to make a complete transition in the whole ecosystem to the new methods and deprecate the old RPC ones 👍

tomaka:
I'm actually not completely happy with this design where the client has to call `storageContinue`.

The function achieves all its goals, but the problem is that the client can delay calling `storageContinue` indefinitely. There exists a limit to the number of active subscriptions, so I'm not worried about DoS attacks or anything like that.

What I'm worried about it a very specific situation: if a load balancer redirects a `storage` query to server A, and later wants to shut down server A because it is no longer needed or needs maintenance, the load balancer currently can generate a `stop` event for all active `follow` subscriptions, wait for all active requests to finish, then shut down server A.

But with the introduction of this `storageContinue` thing, the client can prevent this from happening by keeping a `storage` query active indefinitely.
I feel like there should be something similar to `stop` that the load balancer can generate. The `error` event currently means "fatal irrecoverable error", which isn't the case here as the client should simply try again. I think that `inaccessible` is appropriate, because the load balancer can now longer access the node in question.

tomaka:
I just wanted to mention that I can't merge this pull request myself as I'm not at Parity anymore.

lexnv (approved):
LGTM!

// cc @skunert @jsdw @bkchr

skunert (approved):

src/api/chainHead_unstable_storage.md
```diff
@@ -5,19 +5,28 @@
...
+
+If `type` is `hash` or `descendants-hashes`, then the cryptographic hash of each item is provided rather than the full value. The hashing algorithm used is the one of the chain, which is typically blake2b. This can lead to significantly less bandwidth usage and can be used in order to compare the value of an item with a known hash and querying the full value only if it differs.
+
+If `type` is `closest-ancestor-merkle-value`, then the so-called trie Merkle value of the `key` is provided. If `key` doesn't exist in the trie, then the Merkle value of the closest ancestor of `key` is provided. Contrary to `hash`, a `closest-ancestor-merkle-value` always exists for every `key`. The Merkle value is similar to a hash of the value and all of its descendants together.
```
bkchr, line 26: If I understand it correctly, I find the naming "hard/confusing". Why not "closest-ancestor-merkle-node" and then also saying that this is the hash of the closest merkle node.

bkchr: > I kept the same vocabulary as the spec.

Okay :+1: 

> (the merkle value is either the node value or its hash).

What is the node value? The value attached to a node? Shouldn't we always be interested in the `hash` as we only use this for tracking changes?

tomaka: If the node value is less than 32 bytes than the Merkle value is the same as the node value. If the node value is 32 bytes or more, then it's the hash.
https://spec.polkadot.network/chap-state#defn-merkle-value

bkchr: Ahh okay, ty for the explanation :+1:

PR #41, opened 2023-05-26
Remove `authoringVersion` from runtime spec

### LINKED ISSUE #40: Are all the fields of the runtime spec necessary?
This concern the `spec` field found here: https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_unstable_follow.html#valid

Right now it matches the "runtime version" exposed by the runtime, except that it is missing the `state_version` field.

I was thinking of adding this `state_version`. However, the value of this field is completely useless for JSON-RPC clients. It's an information that the node uses to determine the encoding of the trie.

Rather than add `state_version`, I'm wondering if it doesn't make more sense to leave aside fields that aren't necessary.

We need to leave `transactionVersion` and `specVersion` because they're used to encode transactions. We should maybe also leave `specName`, `implName` and `implVersion` for debugging purposes. `apis` is important to know which entry points are available.

That leaves `authoringVersion` that is useless and could be removed.

tomaka:
Close #40

### DIFF
--- a/src/api/chainHead_unstable_follow.md
+++ b/src/api/chainHead_unstable_follow.md
@@ -175,7 +175,6 @@ The format of the `finalizedBlockRuntime` and `newRuntime` fields can be one of:
     "spec": {
         "specName": ...,
         "implName": ...,
-        "authoringVersion": ...,
         "specVersion": ...,
         "implVersion": ...,
         "transactionVersion": ...,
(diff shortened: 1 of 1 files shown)


tomaka:
I'm still very much in favor of that. Still waiting for a decision.

josepot (approved):
As @tomaka explained in #40: I think that it makes a lot of sense to remove the `authoringVersion` field from the `spec` interface. Also, if this value was needed for debugging purposes, or something else, then this info can be found on the metadata under the `Version` constant of the `System` pallet.

PR #42, opened 2023-05-26
Rename `runtimeUpdates` to `withRuntime`

### LINKED ISSUE #30: chainHead: Calls into runtime without the `runtimeUpdates` flag?
In order to execute `chainHead_unstable_call` on a block provided by the `chainHead_unstable_follow` subscription, the user should provide the `runtimeUpdates` flag set true.
If the `runtimeUpgrades` flag is set, then the substrate will provide additional details to help users identify if a runtime update happened.

Users that are not interested in runtime upgrades could perform `chainHead_unstable_call` directly. Upon a breaking change, the user is responsible to fetch and retry with appropriate parameters.

Considering that updates are not that frequent, what would be the impact of allowing users to call into runtime via `chainHead_unstable_call` regardless of whether the `runtimeUpdates` is set or not?

I'd expect this to have some implications on the light client side. Would it be possible for the light client to execute runtime API calls without having the extra information provided by the `runtimeUpdates` flag?

tomaka:
Close #30

### DIFF
--- a/src/api/chainHead.md
+++ b/src/api/chainHead.md
@@ -12,7 +12,7 @@ _This section contains a small beginner guide destined for JSON-RPC client users
 
 This beginner guide shows how to use the `chainHead` functions in order to know the value of a certain storage item.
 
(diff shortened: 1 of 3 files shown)


lexnv (approved):
LGTM! This sounds more explicit!

// @skunert @jsdw @bkchr

skunert (approved):

PR #43, opened 2023-06-06
Clarifications to the limit to the number of pinned blocks

### LINKED ISSUE #33: chainHead: Soft limit for the number of pinned blocks
Would approx 64 pinned blocks suffice for all light-client interactions?   

The spec is flexible regarding the number of pinned blocks and does not impose any hardcoded values:

> The JSON-RPC server is strongly encouraged to enforce a limit to the maximum number of pinned blocks. If this limit is reached, it should then stop the subscription by emitting a `stop` event. This specification does not mention any specific limit, but it should be large enough for clients to be able to pin all existing non-finalized blocks and a few finalized blocks.

When integrating the pinning API into the `chainhead` methods, we have a limit of 1024 pinned blocks across all subscriptions, coming from substrate.

The short conclusion from https://github.com/paritytech/substrate/pull/13233#issuecomment-1420895166, we are considering:
- soft limit of 64 pinned blocks per subscription (approx 6 minutes - considering all pinned blocks to be finalized)
- a hard limit of 1024 (all available blocks)

After the soft limit is exceeded for a subscription, the subscription could get terminated (via stop event) at any time.

Another idea would be to allow a soft limit of 64 finalized pinned blocks and a soft limit of 128 total blocks; this way we can guarantee ~6 minutes of blocks.

We could even reduce the limits to 32 finalized pinned blocks; and 64 total.  I do remember when we considered the "delay pruning" solution https://github.com/paritytech/substrate/pull/12497 that 32 blocks could be appropriate for a first iteration. 

If that would suffice for all use cases, we could add something similar in the spec to mention the soft limit, while still keeping other implementations of the spec flexible: 

_The substrate JSON-RPC server guarantees 64 finalized pinned blocks and a total of 128 pinned blocks including forks per subscription. After the limit is exceeded, the subscription could be terminated at any time. The user is strongly encouraged to stay within this soft limit. If you need access to older blocks, please use the `archive` methods._

Would love to hear your thoughts on this! 

// CC: @paritytech/tools-team @tomaka @bkchr @skunert @jsdw

tomaka:
Close #33

### DIFF
--- a/src/api/chainHead_unstable_follow.md
+++ b/src/api/chainHead_unstable_follow.md
@@ -151,11 +151,14 @@ The current finalized block reported in the `initialized` event, and each subseq
 
 When a block is unpinned, on-going calls to `chainHead_unstable_body`, `chainHead_unstable_call` and `chainHead_unstable_storage` against this block will still finish normally.
 
-A block is pinned only in the context of a specific subscription. If multiple `chainHead_unstable_follow` subscriptions exist, then each `(subscription, block)` tuple must be unpinned individually. Blocks stay pinned even if they have been pruned from the blockchain, and must always be unpinned by the JSON-RPC client.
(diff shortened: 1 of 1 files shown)


tomaka:
Ping?

josepot (approved):

tomaka:
Reminder that I can't merge my own PRs.

PR #50, opened 2023-06-19
Make `chainHead_unstable_unpin` accept multiple block hashes at once

### LINKED ISSUE #46: RFC `chainHead_unstable_unpin` take multiple hashes
Would it be possible to make the `chainHead_unstable_unpin` call take a list of hashes?

The reason being that more often than not many hashes are unpinned at the same time. Also, since [there are no possible errors](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_unstable_unpin.html#possible-errors) that are specific to hashes, I don't see a reason for not allowing the unpinning of many blocks happen atomically.

tomaka:
Close #46


josepot (dismissed):
LGTM! Thanks @tomaka !

lexnv (dismissed):

PR #51, opened 2023-06-19
Allow querying and returning multiple items at once in `chainHead_unstable_storage`

### LINKED ISSUE #46: RFC `chainHead_unstable_unpin` take multiple hashes

### LINKED ISSUE #47: RFC: add `items` notification for `chainHead_unstable_storage`
Using `chainHead_unstable_storage` with the type `descendants-values` or `descendants-hashes` will in many cases yield thousands of results. Sending an `item` notification for each result creates a lot of unnecessary overhead. That's why I think that it would be a good idea to have an `items` notification for these 2 types.

### LINKED ISSUE #48: missing ability to batch storage requests
When building dApps it usually happens that we must query several storage entries at once.

For instance, let's imagine a situation where the dApp first queries the storage entry of `Session.Validators` in order to query the current set of validators.

Once we have the current validators, we may want to do things like:
- accessing the storage values of `Staking.Validators` for all current validators.
- accessing the storage values of `Identity.IdentityOf` for all the current validators.
- etc

Currently, the JSON-RPC client would have to send a `chainHead_unstable_storage` request per each storage entry. Which is unpractical, it generates a ton of unnecessary overhead and it also slows down by an order of magnitude the time that it takes to get the responses. For instance, currently smoldot "chokes" when the client tries to send so many requests at once, forcing the client to do all sorts of shenanigans in order to get all those results.

Ideally, it should be possible for the client to send many requests in one go, and also [receive the responses in batches](https://github.com/paritytech/json-rpc-interface-spec/issues/47).

I'm not sure what the best API for this should be, though. I have a couple of ideas:

- One idea is to add a valid type `values` into `type` argument and a new `subKeys` argument, for querying a bunch of values from the same dictionary. Then in the `key` the user would pass the key of the dictionary (e.g: for accessing `Staking.Validators` the key would be `0x5f3e4907f716ac89b6347d15ececedca88dcde934c658227ee1dfafcd6e16903`) and then in the individual keys of the dictionary (without prepending the key of the dictionary). For Example:

Querying `Staking.Validators` for the following validators: `16iD4v1FUHMauFB17MsrEFGMrEyGtWEwvgyZ6jCLC5eJdEiZ`, `16ky64Ktu3NWtjrVUZV2RDVe46GvhnUGrxqFrYAQZeB7LHHR` and `16mM7YxjyYqXu7uaJfmfCk9xcr8qqaYUyLazTBt3bueQJAFR` would look like this:

```json
{
   "followSubscription":  0,
   "hash": "0xd484270399fd0ac2e0059704ef231e4da08e038602f71590fd21849ad9ef14fd",
   "type": "values",
   "key": "0x5f3e4907f716ac89b6347d15ececedca88dcde934c658227ee1dfafcd6e16903",
   "subKeys": [
     "0x735e16aa65066f1ffc9ada02baa2ef7a2a24a91f4fdadbbe83a22733dd3ae2deda97f45b4c30dc67",
     "0xa0b1a18158a17b4bfeb586a6d2406aeb67d8dc3895f09dde186e89556ec3c166b6fb1d94bd7a025e",
     "0x19eb0157d3d3179ffeffabd8f3747b5ca19766866bd59856a2e65c537c6456c6f6863c026915f473"
   ]
}
```

- Another option (and probably simplest) would be to allow passing an array of `keys` rather than just one `key` when the type is `values`.

Both options would work perfectly well for my needs, TBH.

I'm a bit worried about how convoluted `chainHead_unstable_storage` can end up becoming with so many overloads and arguments/responses permutations. On one hand it kinda feels like it should be broken down into different (and more specific) functions, on the other hand it kinda makes sense to group all the storage functionality under the same umbrella...

I really don't have any strong opinions about the shape of the final API. The only things that I feel strongly about are:

- It should be possible to send a batch of keys to query many storage entries in one request.
- The addition of the [`values` notification](https://github.com/paritytech/json-rpc-interface-spec/issues/47) (or a notification that serves the same purpose).

tomaka:
Close #48
Close #47

This complicates the API quite a bit, and a lot of subtle details needed to change in order for this to work:

- In the case of a closest descendant Merkle value, we must now provide the original `key` in order to avoid ambiguities when you request the closest descendant Merkle value of multiple items at the same time.
- No error is generated anymore if the trie is empty and you request the closest descendant Merkle value. Instead there's no result now. This complicates a bit the JSON-RPC client code (which is why an error was thrown originally) but is necessary for multiple items to work.
- The server is allowed to return duplicate information, because it might not be possible for the server to de-duplicate information without using tons of memory, while the JSON-RPC client can just filter out whatever it already has.
- The server is free to merge multiple items in the result if it so desired, or leave them split.

### DIFF
--- a/src/api/chainHead_unstable_storage.md
+++ b/src/api/chainHead_unstable_storage.md
@@ -33,7 +33,7 @@ If the `type` of an item is `closest-ancestor-merkle-value`, then the so-called 
 
 If `items` contains multiple identical or overlapping queries, the JSON-RPC server can choose whether to merge or not the items in the result. For example, if the request contains two items with the same key, one with `hash` and one with `value`, the JSON-RPC server can choose whether to generate two `item` objects, one with the value and one with the hash, or only a single `item` object with both `hash` and `value` set. The JSON-RPC server is encouraged to notify as soon as possible of the information at its disposal, without waiting for missing information.
 
-It is allowed (but discouraged) for the JSON-RPC server to provide the same information multiple in the result, for example providing the `value` field of the same `key` twice. Forcing the JSON-RPC server to de-duplicate items in the result might lead to unnecessary overhead.
+It is allowed (but discouraged) for the JSON-RPC server to provide the same information multiple times in the result, for example providing the `value` field of the same `key` twice. Forcing the JSON-RPC server to de-duplicate items in the result might lead to unnecessary overhead.
 
 If a `{"event": "waiting-for-continue"}` notification is generated, the subscription will not generate any more notification unless the JSON-RPC client calls the `chainHead_unstable_storageContinue` JSON-RPC function. The JSON-RPC server is encouraged to generate this event after having sent a certain number of bytes to the JSON-RPC client in order to avoid head-of-line-blocking issues.
 
--- a/src/api/chainHead_unstable_storage.md
+++ b/src/api/chainHead_unstable_storage.md
@@ -4,26 +4,36 @@
 
 - `followSubscription`: An opaque string that was returned by `chainHead_unstable_follow`.
 - `hash`: String containing an hexadecimal-encoded hash of the header of the block whose storage to fetch.
-- `key`: String containing the hexadecimal-encoded key to fetch in the storage.
+- `items`: Array of objects. The structure of these objects is found below.
 - `childTrie`: `null` for main storage look-ups, or a string containing the hexadecimal-encoded key of the child trie of the "default" namespace.
-- `type`: String equal to one of: `value`, `hash`, `closest-ancestor-merkle-value`, `descendants-values`, `descendants-hashes`.
 - `networkConfig` (optional): Object containing the configuration of the networking part of the function. See [here](./api.md) for details. Ignored if the JSON-RPC server doesn't need to perform a network request. Sensible defaults are used if not provided.
+
+Each element in `items` must be an object containing the following fields:
+
+- `key`: String containing the hexadecimal-encoded key to fetch in the storage.
+- `type`: String equal to one of: `value`, `hash`, `closest-ancestor-merkle-value`, `descendants-values`, `descendants-hashes`.
 
 **Return value**: String containing an opaque value representing the operation.
 
-The JSON-RPC server must start obtaining the value of the entry with the given `key` from the storage, either from the main trie or from `childTrie`. If `type` is `descendants-values` or `descendants-hashes`, then it must also obtain the values of all the descendants of the entry.
+For each item in `items`, the JSON-RPC server must start obtaining the value of the entry with the given `key` from the storage, either from the main trie or from `childTrie`. If `type` is `descendants-values` or `descendants-hashes`, then it must also obtain the values of all the descendants of the entry.
 
 The operation will continue even if the given block is unpinned while it is in progress.
 
 This function should be seen as a complement to `chainHead_unstable_follow`, allowing the JSON-RPC client to retrieve more information about a block that has been reported. Use `archive_unstable_storage` if instead you want to retrieve the storage of an arbitrary block.
 
-For optimization purposes, the JSON-RPC server is allowed to wait a little bit (e.g. up to 100ms) before starting to try fulfill the storage request, in order to batch multiple storage requests together.
+`{"event": "items"}` notifications will be generated. Each notification contains a list of items. The list of items, concatenated together, forms the result.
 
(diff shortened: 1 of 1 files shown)


src/api/chainHead_unstable_storage.md
```diff
@@ -149,4 +164,3 @@ No other event will be generated with this subscription.
 - If the `followSubscription` is invalid or stale, then a `{"event": "disjoint"}` notification is generated (as explained above).
 - A JSON-RPC error is generated if the block hash passed as parameter doesn't correspond to any block that has been reported by `chainHead_unstable_follow`.
 - A JSON-RPC error is generated if the `followSubscription` is valid but the block hash passed as parameter has already been unpinned.
-- If the trie is empty and `type` is `closest-ancestor-merkle-value`, then a `{"event": "error"}`.
```
josepot, line 152: Was this line removed by accident? :thinking:

tomaka:
> thus always merging the results on that item

That means that the server might have to wait for some of the results to have come back.
For example, if you ask at the same time for the hashes of the descendants of `0xaa` and for the value of `0xaabb`, the server might find the value of `0xaabb` before it has found the descendants of `0xaa`. If it was forced to merge the results, it would need to keep the value of `0xaabb` on the side for later.

josepot:
> That means that the server might have to wait for some of the results to have come back.
> For example, if you ask at the same time for the hashes of the descendants of `0xaa` and for the value of `0xaabb`, the server might find the value of `0xaabb` before it has found the descendants of `0xaa`. If it was forced to merge the results, it would need to keep the value of `0xaabb` on the side for later instead of sending it back immediately.

I'm afraid that I didn't explain myself very well. With what I'm proposing the server wouldn't have to wait, no. It would just send an Object with what it has ATM. If later it receives more data, then it can send another `items` notification with that data (even if that key has already been seen before). As long as once it's done sending all the data, it  then sends the `done` notification. I don't see why it should have to wait.

But again, if this means a lot of headaches on the server-side, then the API that you are suggesting is good enough, for sure!

josepot:
So, @tomaka: what do you think about emitting a dictionary, rather than emitting an list? Does it make sense? Does it make the spec and/or the implementation on the server too complex?

Once again: that suggestion was just a "it would be even nicer, if...". So, if you are on the fence, then let's go with what's in this PR. The last thing that I would want is for this PR to become stale. Please let me know what's your preference. Thanks!

tomaka:
The server can accommodate both solutions, because if sending an object is too annoying it can always bypass the problem by sending multiple `items` notifications.

However, if the server wants to generate as few `items` notifications as possible, then having to always merge the result forces it to buffer and organize the items that it already has.

In other words, the best possible server-side performance can only be achieved if you return a list.

josepot (dismissed):
@tomaka thank you so much for listening to the feedback and for acting so promptly on it. From the side of someone who is trying to create a JSON-RPC client based on this spec: I have to say that I'm super happy with these changes.

I will let @lexnv merge this, in case that he has some questions or reservations on his end.

src/api/chainHead_unstable_storage.md
```diff
@@ -4,26 +4,36 @@
...
 
-If `type` is `closest-ancestor-merkle-value`, then the so-called trie Merkle value of the `key` is provided. If `key` doesn't exist in the trie, then the Merkle value of the closest ancestor of `key` is provided. Contrary to `hash`, a `closest-ancestor-merkle-value` always exists for every `key`. The Merkle value is similar to a hash of the value and all of its descendants together.
+It is allowed (but discouraged) for the JSON-RPC server to provide the same information multiple in the result, for example providing the `value` field of the same `key` twice. Forcing the JSON-RPC server to de-duplicate items in the result might lead to unnecessary overhead.
```
lexnv, line 36: nit:
```suggestion
It is allowed (but discouraged) for the JSON-RPC server to provide the same information multiple times in the result, for example providing the `value` field of the same `key` twice. Forcing the JSON-RPC server to de-duplicate items in the result might lead to unnecessary overhead.
```

lexnv (dismissed):
LGTM! Thanks for extending this 🙏

PR #52, opened 2023-06-24
Clarify behaviour of `chainHead_unstable_storage`

tomaka:
When you query a `closest-ancestor-merkle-value`, it is possible that the `key` doesn't have any ancestor.

It is not something I have considered, because in practice there's always a trie node whose key is `[]` (empty), and so every key always has an ancestor. But in theory this is not necessarily true. Some chains might not have any trie node at `[]`.

I have additionally clarified the fact that for `value` and `hash` the result doesn't contain any item about a key if the trie doesn't contain a storage value associated to said `key`.

### DIFF
--- a/src/api/chainHead_unstable_storage.md
+++ b/src/api/chainHead_unstable_storage.md
@@ -23,7 +23,7 @@ This function should be seen as a complement to `chainHead_unstable_follow`, all
 
 `{"event": "items"}` notifications will be generated. Each notification contains a list of items. The list of items, concatenated together, forms the result.
 
-If the `type` of an item is `value`, and `key` is associated with a storage value in the trie, then the result will include an item is that contains this storage value. If `key` is not associated with a storage value in the trie, then the result will not include such item.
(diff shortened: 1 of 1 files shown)


src/api/chainHead_unstable_storage.md
```diff
@@ -23,13 +23,13 @@ This function should be seen as a complement to `chainHead_unstable_follow`, all
 
 `{"event": "items"}` notifications will be generated. Each notification contains a list of items. The list of items, concatenated together, forms the result.
 
-If the `type` of an item is `value`, then at least one item in the result is guaranteed to contain the storage value of the `key` of this item.
+If the `type` of an item is `value`, and `key` is associated with a storage value in the trie, then the result will include an item is that contains this storage value. If `key` is not associated with a storage value in the trie, then the result will not include such item.
```
josepot, line 26:
```suggestion
If the `type` of an item is `value`, and `key` is associated with a storage value in the trie, then the result will include an item that contains this storage value. If `key` is not associated with a storage value in the trie, then the result will not include such item.
```

josepot (commented):
Thanks a lot for this @tomaka !

PR #57, opened 2023-06-27
Remove `merkle-value-key` field

tomaka:
Amends #54 

Determining which key corresponds to `merkle-value` requires asking for more data on the networking than necessary otherwise, so I would suggest to remove this field.


tomaka:
After the removal of `merkle-value-key`, it seems that `merkle-value` indicates the Merkle value of `key`. This is misleading.
Consequently I've pushed another commit that renames `merkle-value` to `closest-descendant-merkle-value`.

tomaka:
@josepot @lexnv

josepot (approved):
LGTM!

Issue #23, opened 2022-09-23
Add a minimum to the limit of number of subscriptions


tomaka:
It is impossible to write client-side code that subscribes to things without having a guarantee to the number of subscriptions you're allowed to have.

I think it's important to mention an inferior bound to the limit, such as 64.

PR #60, opened 2023-07-16
Stabilize `transaction`-prefixed functions

tomaka:
One remark concerning https://github.com/paritytech/json-rpc-interface-spec/issues/23 is that the `dropped` event can be used if the JSON-RPC client has reached its maximum number of subscriptions. This is already explained in the documentation of the function. There is therefore no need to change anything.


Polkadot-Forum:
This pull request has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/new-json-rpc-api-mega-q-a/3048/14

jsdw:
Will close in favour of #144

PR #61, opened 2023-07-17
Remove `chainHead_unstable_genesisHash`

tomaka:
Since the objective is to stabilize the `chainHead` namespace in the nearby future, this pull request proposes to remove the `chainHead_unstable_genesisHash` JSON-RPC function out of caution.

The reason is that a client that simply follows the head of the chain might not necessarily be aware of what the genesis hash is. See https://github.com/polkadot-fellows/RFCs/pull/8 for an example.
The genesis hash can be found in the chain at `System::BlockHash[0]`, but since this is a runtime-specific thing it should be implemented by the JSON-RPC client on top of `chainHead_storage`, rather than by the server.

Note that `archive_genesisHash` and `chainSpec_genesisHash` still exist, as a node that implements `archive` is clearly supposed to know what the genesis hash is, and that a node that implements `chainSpec` can read the genesis hash from its chain spec.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -20,7 +20,6 @@
     - [chainHead_unstable_body](api/chainHead_unstable_body.md)
     - [chainHead_unstable_call](api/chainHead_unstable_call.md)
     - [chainHead_unstable_follow](api/chainHead_unstable_follow.md)
-    - [chainHead_unstable_genesisHash](api/chainHead_unstable_genesisHash.md)
     - [chainHead_unstable_header](api/chainHead_unstable_header.md)
     - [chainHead_unstable_stopBody](api/chainHead_unstable_stopBody.md)
     - [chainHead_unstable_stopCall](api/chainHead_unstable_stopCall.md)
--- a/src/api/chainHead_unstable_genesisHash.md
+++ b/src/api/chainHead_unstable_genesisHash.md
@@ -1,9 +1,0 @@
-# chainHead_unstable_genesisHash
-
-**Parameters**: *none*
-
-**Return value**: String containing the hexadecimal-encoded hash of the genesis block of the chain.
-
-This function is a simple getter. The JSON-RPC server is expected to keep in its memory the hash of the genesis block.
-
-The value returned by this function must never change for the lifetime of the connection between the JSON-RPC client and server.


josepot:
Probably a silly question, but what is the point of having both `archive_genesisHash` and `chainSpec_genesisHash`? wouldn't it make sense to remove `archive_genesisHash` as well? Or this is it perhaps because archive nodes are not meant to implement the `chainSpec` API?

tomaka:
The point is that the JSON-RPC client doesn't know which API is available. Some nodes might have `archive` but not `chainSpec`, or vice versa, or neither, or both.

jsdw (approved):
Sounds fair to me!

Side question: can we not keep eg `chainHead_genesisHash` unstable even if we stabilise the rest of the interface?

More generally, if we would stabilise the individual calls that we were happy about on a case by case basis, then the path to complete stabilisation may be quicker, and it would allow us to "experiment" with unstable methods that we weren't sure about in the meantime (or add unstable methods later etc that may or may not be stabilised).

lexnv (approved):

jsdw:
Ok that makes sense! So if we want to add new methods, they can take this sort of path I guess?:

- `chainHead_v1` methods are stable
- We add a new `chainHead_unstable` method or methods that we think might be useful.
- At some point we want to stabilise those, but the `v1` interface is fixed, so we expose all of the `v1` methods now as `chainHead_v2` and add the unstable methods to the `v2` namespace at the same time.
- We can keep the "original" `v1` methods also exposed in the `v1` namespace for backward compat (seems sensible to do this for as long as possible)

tomaka:
> Ok, thanks for the clarification, that makes sense! So if we want to add new methods, they can take this sort of path I guess?:
> 
>     * `chainHead_v1` methods are stable
> 
>     * We add a new `chainHead_unstable` method or methods that we think might be useful.
> 
>     * At some point we want to stabilise those, but the `v1` interface is fixed, so we expose all of the `v1` methods now as `chainHead_v2` and add the unstable methods to the `v2` namespace at the same time.
> 
>     * We can keep the "original" `v1` methods also exposed in the `v1` namespace for backward compat (seems sensible to do this for as long as possible)

Yes! I'm making the bet that this shouldn't happen too often, because missing a function indicates either a big fuck up (which normally shouldn't happen) or that some design aspect of Substrate/Polkadot has changed (which shouldn't happen very often).

jsdw:
Thanks, that makes sense!

@josepot FYI I'm happy for this to merge when you are!

(also FYI, the `chainSpec_v1_` stable methods are already implemented and exposed in Substrate at least in case that helps, so we have a stable way to obtain genesis hash from that namespace)

josepot (approved):
Thanks!

josepot:
> Thanks, that makes sense!
> 
> @josepot FYI I'm happy for this to merge when you are!
> 
> (also FYI, the `chainSpec_v1_` stable methods are already implemented and exposed in Substrate at least in case that helps, so we have a stable way to obtain genesis hash from that namespace)

Sorry, I thought that I had already approve this 😬

PR #65, opened 2023-07-19
fix: `inaccessible` event on `chainHead_unstable_call`

josepot:
I'm pretty sure that this was a copy/paste mistake as non of the other `inaccessble` events have the `error` property. Also, IMO it doesn't make a lot of sense to have an error property on this event.

cc: @tomaka

### DIFF
--- a/src/api/chainHead_unstable_call.md
+++ b/src/api/chainHead_unstable_call.md
@@ -56,14 +56,11 @@ No more event will be generated with this `subscription`.
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


lexnv (approved):

tomaka (approved):
Either that or adding an `error` to the other calls works for me

josepot:
> Either that or adding an `error` to the other calls works for me

I think that in this case, the `event` property already provides enough info. So, I would rather remove the `error` prop.

PR #66, opened 2023-07-20
Change the `chainHead` functions to have proper limits

### LINKED ISSUE #22: Clarify how the server handles a client reaching its subscriptions count limit
It should be handled by sending back `"dropped"` events.

### LINKED ISSUE #23: Add a minimum to the limit of number of subscriptions

### LINKED ISSUE #58: `chainHead_unstable_storage` needs a limit to the number of items
cc https://github.com/paritytech/json-rpc-interface-spec/pull/51

I now remember why prior to https://github.com/paritytech/json-rpc-interface-spec/pull/51 each call to `chainHead_unstable_storage` could only subscribe to one item.
The reason is that we need a limit to the number of items you can subscribed to at any given time, as otherwise an attacker could subscribe to millions of keys.

### LINKED ISSUE #63: `chainHead_storage`: Extend reported error events
The [chainHead_storage](https://github.com/paritytech/json-rpc-interface-spec/blob/main/src/api/chainHead_unstable_storage.md) spec mentioned the following about errors:

```bash
{
    "event": "error",
    "error": "..."
}
```

> error is a human-readable error message indicating why the call has failed. This string isn't meant to be shown to end users, but is for developers to understand the problem.
This can only be the first event generated by this subscription. No other event will be generated with this subscription.

Considering that we may have multiple items to query, I do wonder:
- Could we generate the error event after providing items responses?
  We might receive a query [Value::A, Value::B, Hash::C]. Processing the items we would produce in the following order:
    - `event = items with value::A`: provides the storage value for A
    - `event = error key=0x.. type=Value`: failed to fetch from storage for some reason (ie OOM, bad disk sector)
    - `event = items with hash::C` : provides the storage hash for C
    - `event = done`: this transforms the events a bit a makes the error event non-terminal since we could and may process other items

- We would need a way to identity which error is coming from which item in the provided query. We could currently encode the `query::key` and `query::type` into the error string but that might be a bit unfriendly. Would an error of the following shape provide more useful info?

```bash
{ "event": "error", "key": "0x..", "type": Hash, "error": "String with reason" }
```

Such we could disambiguate between multiple query types made with the same query key, and would allow the server to process items out of line in arbitrary order.

The same would apply for `inaccessible` as well
> The inaccessible event indicates that the storage value has failed to be retrieved from the network.
No more event will be generated with this subscription.

// cc: @tomaka @josepot @jsdw @paritytech/subxt-team 

Related to: https://github.com/paritytech/substrate/pull/14551

tomaka:
Close #58 
Close #23 
Close #22 
Close #63

See discussion in #58 

The list of changes are:

- `chainHead_storage`, `chainHead_body`, and `chainHead_call` are no longer subscriptions but simple request-responses.
- `chainHead_storage`, `chainHead_body`, and `chainHead_call` now return a JSON object rather than just a `subscriptionId`. This JSON object indicates whether the operation could be successfully started.
- The notifications that were previously generated by `chainHead_storage`, `chainHead_body` and `chainHead_call` are now generated by the `chainHead_follow` instead, and now have an `operation-` prefix.
- `chainHead_storageContinue` has been renamed to `chainHead_continue` and must now be passed a `followSubscriptionId`.
- `chainHead_stopBody`, `chainHead_stopStorage` and `chainHead_stopCall` have been merged into `chainHead_stopOperation` and must now be passed a `followSubscriptionId`.
- In the case of `chainHead_storage`, when a subscription successfully starts, is also indicated the number of items that were ignored. Note that it is possible for all items to be discarded, in which case the subscription will just end with `"event": "done"` right after.
- Removed `networkConfig` from everywhere. This turned out to be too difficult to implement for the server when it needs to manage its own resources.
- Removed the `disjoint` events from `chainHead_storage`, `chainHead_body`, and `chainHead_call`, and replace them with a `limitReached` return value. Even though this situation is more appropriate as a return value, it was previously an event in order to make the API easier to use. Since we're switching to returning an object instead of just a `subscriptionId`, might as well handle `disjoint` the proper way as well.
- `chainHead_follow` might now return a JSON-RPC error if there are already more than 2 active subscriptions.
- In the case of `chainHead_storage`, it is no longer forbidden to generate an `error` event after some `items` notifications have been generated.

Note that `error` is still events rather than a return value. In the case of `chainHead_unstable_call`, a light client must first download a bunch of stuff, and then only later an `error` might happen, and thus it is more appropriate as an event.
In the case of `chainHead_unstable_storage`, the `error` could be a return value, but I kept it as an event for consistency.

## Remarks for JSON-RPC client developers

When it comes to `chainHead_follow`, there is a minimum of two active `chainHead_follow` subscriptions per client. **This means that you normally don't need to handle a situation where a JSON-RPC error is returned.**

When it comes to storage requests, body requests, and call requests, the requests to start should be treated as if they were a queue. If `limitReached` is returned or `discardedItems` is non-zero, just pause the queue. If the queue is too large, you **must** either discard unnecessary entries or show error messages to the user. Maybe do read about queue theory on the Internet to understand what this is about.


josepot (approved):
The refactor on my end is going to be quite significant 😅 . However, I'm convinced that this is a much better approach. Thanks @tomaka !

src/api/chainHead_unstable_body.md
```diff
@@ -4,9 +4,50 @@
...
+The JSON object returned by this function has one the following formats:
+
+### Started
+
+```
+{
+    "result": "started",
+    "subscriptionId": ...
+}
+```
```
lexnv, line 19: I do wonder if the `subscriptionId` from here is the same ID of the subscription or some other value we could generate?

For example to subscribe to a producing method using the JsonRPSee we'd have to:
- submit a ` {"id": 1, "method": "chainHead_unstable_body", "params": [..]}`
- the server would then give us back: `{"jsonrpc":"2.0","id":1,"result":"0xb0d1..."}`
  - the result is a string which is the `subscriptionId` 
- then the server would start producing events of the following format
```json
{"jsonrpc":"2.0","method":"chainHead_unstable_body","params":{
     "subscription":"0xb0d1...",
     "result":{ 
          "result": "started",
          "subscriptionID": "what-should-be-placed-here?"
       }
   }
}

```

src/api.md
```diff
@@ -8,28 +8,3 @@ Any missing parameter, or parameter with an invalid format, should result in a J
 
 - "hexadecimal-encoded" designates a binary value encoded as hexadecimal. The value must either be empty, or start with `"0x"` and contain an even number of characters.
 - "SCALE-encoded" designates a value encoded using [the SCALE codec](https://docs.substrate.io/v3/advanced/scale-codec/).
-
-## The `networkConfig` parameter
```
lexnv, line 12: Nice! With this, we'd simplify code from substrate's point of view! 👍

src/api/chainHead_unstable_body.md
tomaka: I don't understand the question?
The server would give you back `{"jsonrpc":"2.0","id":1,"result":{"result": "success", "subscriptionId": "0xb0d1..."}}`

lexnv: For the pubsub model of subscription based notifications, I believe we'd expect first:
- `{"jsonrpc":"2.0","id":1,"result":"0xb0d1..."}`which has the subscription ID
- then a notification that has the prefix:
```json
{"jsonrpc":"2.0","method":"chainHead_unstable_body","params":{
     "subscription":"0xb0d1...",
     "result":{ 
        [event here]
```

Would this modify the format of pubsub? (we might need to do some changes to the jsonrpsee IIUC) 

The jsonrpsee would expect only a subscription ID to be present in the `result` body as a string.
From looking at our code at [code/src/server/rpc_module.rs](https://github.com/paritytech/jsonrpsee/blob/f7469b9ae2e9bc2f1260b8a75028cfc4b337ed39/core/src/server/rpc_module.rs#L373-L393).

// cc @jsdw to make sure I've got this right

tomaka: > Would this modify the format of pubsub?

Well, yes, but "pubsub" is not any kind of standard that exists. If jsonrpsee assumes that all functions that start a subscription return just a `subscriptionId`, it's their fault.

The alternative is to return a `subscriptionId`, but then the subscription starts in a "not actually started" phase, and would then generate a `"started"` event as soon as possible. This would work but would be a stupid API.

lexnv: Yep, indeed, we could extend the jsonrpsee server to parse other information than the subscription ID for these methods.

Would it be possible to presume with the pubsub format the following?
- when the server provides a result `{"jsonrpc":"2.0","id":1,"result":"0xb0d1..."}`, then the method is in `started` state
- for `limitReached` and `disjoint` as initial operations, the server would respond with a JSON-RPC error object 

There might still be a few things we want to tick of in JsonRPSee before being able to do a proper merge into substrate and the different format support might be blocked for a bit. Also, it would probably mean a breaking change from the user's perspective that we have accommodated to the pubsub

tomaka: > Would it be possible to presume with the pubsub format the following?

>    when the server provides a result {"jsonrpc":"2.0","id":1,"result":"0xb0d1..."}, then the method is in started state
>    for limitReached and disjoint as initial operations, the server would respond with a JSON-RPC error object

The reason why I made this change is for `chainHead_storage`, where the server might both successfully start the subscription AND discard some items.

jsdw: > There might still be a few things we want to tick of in JsonRPSee before being able to do a proper merge into substrate and the different format support might be blocked for a bit. Also, it would probably mean a breaking change from the user's perspective that we have accommodated to the pubsub

As I understand it, the JSON-RPC spec doesn't contain anything about subscriptions, and so the JSON-RPC impl in ethereum added this approach of returning  ` { "result":"0xb0d1..." }` to indicate a subscription had started, and then `{ "result": { "subscription": "0xb0d1...", params: ...}  }` for ongoing subscription responses. Substrate continued this approach.

We could change this, but it would be a deviation from this "unofficial standard" and from how Etherum does things, and would require all JSON-RPC client impls that support this (Polkadot.js/CAPI included) to be changed. So it has to be worth this breakage.

If we do go ahead with the change, `jsonrpsee` (and probably other clients) would need to support both old and new formats being used side by side, at least until the new methods are stabilised and generally available, and potentially forever if we want to continue to support its usage outside of the Polkadot ecosystem. Ie, we'd want to handle both `"result":"0xb0d1..."` and `"result": { "subscriptionId": ""0xb0d1...", "result": ...}`.

One approach to do this in `jsonrpsee` would be for it to try and decode both formats to find which one works. Maybe we can add some annotation on the proc macro stuff or runtime flag could be used to toggle between them per call. @niklasad1 will likely have opinions on this, so let's see what he thinks when he's back from holiday (week after next IIRC).

tomaka: > We could change this, but it would be a deviation from this "unofficial standard" and from how Etherum does things, and would require all JSON-RPC client impls that support this (Polkadot.js/CAPI included) to be changed. So it has to be worth this breakage.

I'm in general very skeptical of the model where JSON-RPC clients automatically detect which subscription an incoming notification belongs to and redirect it to a specific callback.
I think that this model creates some spaghetti of callback hell (something JavaScript can very easily fall into).

It worked with the legacy JSON-RPC API, because it was designed to be as close to the UI as possible. If the UI needed something, there was a JSON-RPC function specifically for that need.

However, the new JSON-RPC API requires more code on the client side than the old API, and no matter the outcome of the discussion I'm replying to, I have huge doubts that it's possible to write sane code with a model like that.

I'm pretty convinced that it's better if JSON-RPC clients do *not* try to automatically parse subscription-related messages, and instead simply update a single state machine synchronously.

josepot: > As I understand it, the JSON-RPC spec doesn't contain anything about subscriptions

You're right, the JSON-RPC specification doesn't directly provide for subscriptions. However, it does include a concept of "Notifications". These are essentially one-way messages (without an `id`) that don't anticipate a reply. With this in place, we can construct our own conventions to create a pub/sub model.

While there isn't a universally agreed-upon way to build pub/sub on top of JSON-RPC, I believe we can establish a convention that meets our specific needs. We've previously ventured into unconventional territory with our spec, as in the case of `chainHead_unstable_follow`, where the generated subscriptionId has to be used for all other `chainhead` calls. This idea of "nested" subscriptions isn't common, so using a library built for the "unofficial standard" of "how Ethereum does things" might present challenges as subscriptionIds are usually hidden (b/c they are deemed as implementation details).

> and would require all JSON-RPC client impls that support this (Polkadot.js/CAPI included) to be changed

I don't foresee Polkadot.js adopting this new spec. But as for CAPI, I've been developing a library to simplify using this new JSON-RPC API. This library will be standalone and open for any other project to use or improve, including Polkadot.js.

> If we do go ahead with the change, jsonrpsee (and probably other clients) would need to support both old and new formats being used side by side

From my perspective, you could use jsonrpsee for APIs built on the "ethereum convention" and the "CAPI library" for APIs following the "substrate convention". This way, the two could work in parallel rather than expecting all clients to support both old and new formats simultaneously.

jsdw: In terms of the API, I've been reading through the changes and I'm generally onboard with the idea that subscription responses are changed from `"result": "0xb0d1..."` to `"result": { "subscriptionId": ""0xb0d1...", "result": ...}` from a DX perspective. Ultimately, it allows 

I do care that there is always a consistent location for the "subscriptionId" (whether it be `response.result` or `response.result.subscriptionId`), so that client libraries can continue to provide a nice "high level" interface like:

```rust
fn subscribe(&self, method_name) -> Stream<Item>
```

> I'm pretty convinced that it's better if JSON-RPC clients do not try to automatically parse subscription-related messages

I can see that perspective (and if I were writing a JSON-RPC library today from scratch with it in mind, I'd likely provide the interface above but also a lower level "stream of all notifications from server" that could be used instead when it made sense).

> I'm in general very skeptical of the model where JSON-RPC clients automatically detect which subscription an incoming notification belongs to and redirect it to a specific callback.
I think that this model creates some spaghetti of callback hell (something JavaScript can very easily fall into).

I'm not really sure why the example interface above would lead to spaghetti code or whatever myself in most cases (it's worked well for us in Subxt I think so far), but being able to "opt out" in certain cases where it would does make sense to me. In any case, I expect most JSON-RPC client implementations would wish to expose something like the above example, but as I understand it this will still be possible with the proposed change anyway.

> I don't foresee Polkadot.js adopting this new spec.

It'll have to eventually if we are planning to remove the old APIs (granted, that will likely be a while away though). As will any third party JSON-RPC libraries that are currently in use across the ecosystem.

-------

Anyway, I do think that this change comes with costs:

- Getting this update implementedf in Substrate will take longer: 
  - We would need to adapt `jsonrpsee-server` in Substrate to be able to return arbitrary data along with subscription IDs for the new APIs, but have it continue working as it does for the old ones. @niklasad1 will no doubt have thoughts on this all when he's back.
  - Logistically, we need to merge some in flight backpressure stuff in `jsonrpsee-server` before we can make these changes.
- Adoption of the new APIs may take logner too:
  - We'd need to adapt `jsonrpsee-client` or write a new JSON-RPC client library for Subxt. 
  - Moreover, _everybody else in the ecosystem_ would need to do adapt or migrate to RPC clients that support this change.

Are we happy paying them?

tomaka: > Are we happy paying them?

Another thing to consider is that if we go for the clumsy but easier to implement API, I'm pretty convinced that we will never be able to switch to a nicer API. There's going to be an issue titled "make the API less clumsy" opened in this repository, and because it requires a lot of changes to implement, nobody will ever care about it and the issue will rot forever.

tomaka: Another possible version of the API could be:

- `chainHead_storage` becomes a simple request RPC (no longer a subscription), and returns the number of items that have been queued.
- Later, when the storage item has been retrieved, it's `chainHead_follow` that produces the notification.

This is clean, but is quite a big change in the spec.

jsdw: To add to `fn subscribe(&self, method_name) -> Stream<Item>`, I do see the perspective that everything is just a standard JSON-RPC method call, but some of those responses may also provide a "subscriptionId" and you're welcome to use that to associate incoming notifications with a particular call if you like.

But I also do think that its useful if a client is able to:
- Automatically tie subscription responses to a single `Stream` or whatever so that the user doesn't have to. 
- Provide a nice interface to end the subscription (by calling the related "unsubscribe" method or what have you under the hood).

These do require that there is still uniformity in how subscriptions and unsubscriptions are handled, and I do want to make sure that we keep that.

tomaka: > But I also do think that its useful if a client is able to:

    Automatically tie subscription responses to a single Stream or whatever so that the user doesn't have to.
    Provide a nice interface to end the subscription (by calling the related "unsubscribe" method or what have you under the hood).

My post yesterday is precisely that I don't agree with that. With the new JSON-RPC API, these features hurt you way more than they're helping. I'm telling this from experience having written a client-side library that uses the new JSON-RPC API.

jsdw: > My post yesterday is precisely that I don't agree with that. With the new JSON-RPC API, these features hurt you way more than they're helping. I'm telling this from experience having written a client-side library that uses the new JSON-RPC API.

Yeah, I get that! But I would say that it's still worth being consistent with how subscriptions are handled even if the Stream abstraction proves useless, just because its easier to document/teach/understand them that way.

> Another possible version of the API could be:
>    - chainHead_storage becomes a simple request RPC (no longer a subscription), and returns the number of items that have been queued, and maybe some kind of identifier for the operation.
>    - Later, when the storage item has been retrieved, it's chainHead_follow that produces the notification.
>    - Same for chainHead_body and chainHead_call.

If you think that this is as clean/better than the current suggestion, and it didn't lead to changing the subscription protocol (and so less churn), and it's not likely that some other compelling reason to change the subscription protocol will emerge, then my 2c is that it would be an overall win :)

josepot: > I'm telling this from experience having written a client-side library that uses the new JSON-RPC API.

Speaking from my own experience of crafting a client-side library for the new JSON-RPC API, I'm totally with @tomaka on this one.

Please @jsdw [check this out](https://stackblitz.com/edit/stackblitz-starters-j52yw9?file=src%2Fmain.ts). It's a sneak peek at a "low-level" library I've been cooking up to interact with the new JSON-RPC API. Feel like exploring it further? Just [clone this repo](https://github.com/josepot/test-new-json-rpc-api) and dig in.

This library doesn't make assumptions for you, like unpining blocks—that's on you. But it does provide a handy, strongly-typed API for key interactions with the new JSON-RPC API.

CAPI plans to use this as a base library to built upon, adding some opinions to better cater to their needs. They'll automate things like unpining blocks, and roll out some opt-in user-friendly APIs (like observables, async iterators, etc).

> However, please notice how despite the fact that library is fairly low level, the changes that are being discussed in this discussion are not going to affect its public API.

Just a heads up—despite the library being low level, the changes we're talking about here won't mess with its public API. They're behind-the-scenes tweaks that won't affect you.

I'd be super happy to make a companion library in Rust, so subxt could also benefit from a library like this.

Here's the thing: there's no set standard for building pub/sub models over JSON-RPC, so why not create conventions that suit us? As long as we can whip up basic APIs to hide the nitty-gritty details, we're fine.

jsdw: > Here's the thing: there's no set standard for building pub/sub models over JSON-RPC, so why not create conventions that suit us? As long as we can whip up basic APIs to hide the nitty-gritty details, we're fine.

I agree, and we should totally break things when it leads to better stuff. But the other side of the coin is; why abandon existing conventions that have been used in these parts for years if there is no need to. So for me it's simply a case of; is there a need?

If we can have clean APIs that stick to the current protocol (eg https://github.com/paritytech/json-rpc-interface-spec/pull/66#discussion_r1270593268?), save a bunch of churn and maintain compatibility with existing clients (inc the `jsonrpsee-server` lib used in Substrate to provide these methods) then that'd be my preference.

If this breaking change will actually enable better APIs and such then of course we can put the work in to make the relevant changes in jsonrpsee/Substrate etc to make it happen, accept that there will be churn in the ecosystem, and ultimately I'll get behind that.

> I'd be super happy to make a companion library in Rust, so subxt could also benefit from a library like this.

Ultiamtely if we go this route, I'd prefer that our team looks into this as I'd make something with Subxt's needs/RPC interface in mind and probably alter said interface a bit in the process, but thankyou for the offer :)

josepot: > If we can have clean APIs that stick to the current protocol

Even we went ahead and did it like in [#66 (comment)](https://github.com/paritytech/json-rpc-interface-spec/pull/66#discussion_r1270593268) we wouldn't be sticking the "current protocol", because we have "nested subscriptions".

I'm pretty sure that the "current protocol" (whatever that is, b/c it's an ad hoc convention which doesn't have any spec behind) doesn't account for the fact that a "subscriptionId" can be used for anything else than listening to the messages of that subscription.

jsdw: > Even we went ahead and did it like in https://github.com/paritytech/json-rpc-interface-spec/pull/66#discussion_r1270593268 we wouldn't be sticking the "current protocol", because we have "nested subscriptions".

The actual protocol remains exactly the same though (ie `"result": id` returned for subscription init and `{ "subscription": id }` for subscription messages); the only thing that changes is that the client would need to expose the subscriptionId to the user so that the user could make use of it in other calls, which is a very small thing and just an API detail.

josepot: > The only thing that changes is that the client would need to expose the subscriptionId to the user so that the user could make use of it in other calls, which is a very small thing and just an API detail

Sure, let me help clarify your points:

The crux of my argument is that developing an API around implementation details is a bad idea. This ad hoc convention is just that - an implementation detail.

In my view, if a library exposes an API centered around these implementation details, it's usually a better idea to craft a new library that conceals these technicalities and focuses on presenting a more helpful API. This is generally better than trying to contort the existing library to accommodate a new twist in this ad hoc convention. [Duplication is far cheaper than the wrong abstraction](https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction).

I find it puzzling to request a major change to the spec based on an ad hoc convention that was used in the old JSON-RPC API. This is especially the case when one of the main drivers for creating this spec was to shed that old baggage.

Moreover, your decision on what constitutes "a very small thing and just an API detail" seems a bit arbitrary to me. In my opinion, the same could be said about a change accommodating this API: you'd simply need to include an optional mapper function (aka selector). The default value for this would be an identity function to select the value of the subscriptionId.

josepot: Regardless, if @tomaka finally decides to roll with the [option put forward in this comment](https://github.com/paritytech/json-rpc-interface-spec/pull/66#discussion_r1270593268), I've got one little plea: please ensure there's a mechanism to cancel an ongoing request 🙏. This is especially important considering that requests can take a while to resolve and might occupy a bunch of "slots" that aren't needed anymore.

jsdw: > The crux of my argument is that developing an API around implementation details is a bad idea. This ad hoc convention is just that - an implementation detail.

It's not been formally specified, but the current protocol around handling subscriptions is much more than an "ad hoc convention" IMO; it's how all JSON-RPC clients here and in Etherum have agreed on representing the notion of a subscription. It's the de facto standard.

> Moreover, your decision on what constitutes "a very small thing and just an API detail" seems a bit arbitrary to me. In my opinion, the same could be said about a change accommodating this API: you'd simply need to include an optional mapper function (aka selector).

Exposing the "subscriptionId" from the client library to the user doesn't change anything about the protocol (ie how subscriptions are represented on the wire). So yes, I do think that it is an API detail and not a protocol detail. From my perspective, the change to expose this was trivial, but the change to subscriptions (both the subscriptionId now living in an object, and the other details of that object now also being relevant) will be far less trivial to adapt to (I'm thinking as much about the server code in Substrate as any client library here).

Anyway, we could go back and forth on this forever, and I don't wish to do that. If the decision is ultiamtely to change the way we return subscription things then that's fine, and I'll get behind it. The pros and cons have already been discussed.

josepot: For context, I've created and currently maintain an Ethereum JSON-RPC client, so I'd like to chime in with a few points:
1) The [official JSON-RPC Spec](https://ethereum.github.io/execution-apis/api-documentation/) doesn't touch on subscriptions because it's designed as a pull-based API spec, tailored for use with traditional RPC nodes.
2) It's also true that there's a [widely accepted convention](https://geth.ethereum.org/docs/interacting-with-geth/rpc/pubsub) for Ethereum clients that do support subscriptions.
3) This specific convention within Ethereum isn't just about returning an opaque ID in the response result. It also specifies that all subscription and unsubscription actions are carried out using two JSON-RPC methods: `eth_subscribe` and `eth_unsubscribe`. Furthermore, it sets rules for how these methods should be used.
4) Ethereum client libraries that adhere to this convention usually handle it through polling, as most Ethereum providers don't allow `eth_subscribe` and `eth_unsubscribe` calls. This is just to highlight the fact that we are in a completely different situation.

While I personally struggle to see why we should aim for compliance with this convention, if we decide to go down that path, let's at least do it right. Let's have two dedicated methods for handling all subscriptions and unsubscriptions, and let's make sure we properly document this convention.

jsdw: I think something that's maybe grating on me a bit with this change is also that I think the current convention is quite clean IMO.

Right now, when you call a subscription method, iirc you get either:

- `"result": subscriptionId` - subscription worked; I know this because I got "result" back. The result is always an ID to associate future msgs with.
- `"error": { error object }` - an error occurred so the no subscription will occur.

I think that's fairly clean. (This V2 API Might not emit any errors actually; can't recall)

With the proposed change, calling a subscription method would give:

- `"result": { "subscriptionId": id, "foo": ..., "bar": ...}` - since I can deserialize "subscriptionId", I can see a subscription will occur. Some other props in the same object may come back too to be handled.
- `"result": { "foo": ..., "bar": ...}` - I'm unable to deserialize "subscriptionId", so I know a subscription will not occur. Some other props in the same object may come back too to be handled.
- `"error": { error object }` - an error occurred so the no subscription will occur. (maybe this is unused here though).

To me, that feels more complicated to work with/understand. I have to parse more to see if I am to expect follow-up messages. I have other props as well to deserialize in there. The client is pushed towards working at a lower level (ie not having built in subscription logic) because more efficient to deserialize all props inc subscriptionId once and then logic asking for type would check for subscriptionId and handle it (in JS this is easier than in Rust).

Maybe you guys think the latter is actually cleaner. 

It probably comes down to a difference in philosophy around whether a client library should handle subscriptions or whether to push that higher up. 
- Because it's been a useful abstraction in the past, I like the idea that the protocol is simple enough that the client library can handle it fairly easily, as now. 
- Coming from a perspective of the client not caring about subscriptions, there becomes no need to have any convention around where, or if, a "subscriptionId" is returned; it's just a response. The highert level can interpret the response and then interpret notifications accordingly.

So just to say, I do think I understand where you are coming from (but correct me if wrong). Whether it's a good way to head or not, I am still very on the fence about. Till now, it's been fairly easy to envisage updating Substrate and Subxt etc to support the new API, but this more fundamental shift in thinking would also require more fundamental changes in order to best suit the new philosophy.

josepot: Hey, I promise this is my last attempt to clarify my point of view, and I really appreciate your patience with this 🙂

Firstly, I understand where you're coming from. However, please notice that the convention you're referring to was tailor-made for Ethereum. It was specifically designed to allow a pull-based API to pretend to be a subscription-based one using polling on the official Ethereum JSON-RPC API. The simplicity and limitations of this method (just supports four operations) is exactly what makes it possible to simulate through polling, especially for Ethereum providers who don't support these operations natively. It's a specific solution for a specific challenge that Ethereum was facing.

In the context of our "legacy" Substrate JSON RPC API, parts of this convention were adopted, mainly because "that's how Ethereum does it". I totally get why that decision might have been made at the time, but it's important to note that Ethereum had its own specific reasons for doing things the way they did. The thing is, we've adopted some parts of the convention while leaving out others (we have many methods scattered around for subscribing/unsubscribing), leading to the creation of certain libraries and abstractions that fortify this altered version of the Ethereum convention.

But let's imagine a world where you've never come across this convention and no abstractions have been built around it. All you're familiar with is the [JSON-RPC 2.0 Spec](https://www.jsonrpc.org/specification) - which basically just defines requests/responses and notifications.

Now, looking at it from this perspective, let's revisit our problem: we need to request a bunch of storage items, and we want the response to contain two things - the number of items the server was able to handle, and the subscriptionId for listening to future notifications. If we could design this from scratch, having both of these in the response seems pretty intuitive, right? We're calling an async function and it'd be awesome if this function could return a tuple containing exactly what we need.

However, you've suggested that it's not "clean" because it doesn't adhere to the previously-mentioned convention, which is centered around a different problem and restricts us to only returning the subscriptionId, not a tuple. To align with this convention, you're even open to a major spec rewrite. 

Take a look at my [initial comment](https://github.com/paritytech/json-rpc-interface-spec/pull/66#pullrequestreview-1538859738), which I posted before our conversation began. Why was I talking about significant refactoring? It's because I too had a bunch of code based on the assumption that a method returning a subscriptionId would always find it in the `result` property. Coming to terms with this new approach wasn't easy for me either, but I eventually realised that it would be worth it.

Even though the refactoring can make the internals a bit more complex, the end result will be a library that -while preserving the same public API- behaves better - it won't error out when there are too many ongoing requests and it'll process all requests efficiently. This way, the consumer won't even notice if a limit is reached. So yes, the inside workings may be a bit more complex, but that's the essence of building a library: to manage complexity internally so that users don't have to. 

I know this is a big change and it might not be easy to accept, but I genuinely believe it'll be for the best. The main goal is to create a spec that allows us to handle our needs as straightforwardly as possible, and I feel that adhering strictly to the old convention might not be the most helpful approach here.

tomaka: > It's not been formally specified, but the current protocol around handling subscriptions is much more than an "ad hoc convention" IMO; it's how all JSON-RPC clients here and in Etherum have agreed on representing the notion of a subscription. It's the de facto standard.

It's only a de facto standard in the blockchain ecosystem, which is filled with extremely technically weak project which just copy paste each other.
And just because it's a de facto standard does in no way mean that it's good.
To give a similar example, when you start an Ethereum 2 node, it also starts an Ethereum 1 node in the background. It's not for any good reason, but because it was too hard in terms of engineering to merge the Ethereum 1 and 2 codes in one. In 5 years people might argue that it's a de facto standard to split your node into two processes that send messages to each other, but that still doesn't make it right.

> because more efficient to deserialize all props inc subscriptionId once

Let's not use "efficient" as an argument. We're talking about messages that are at most a few hundred bytes long. It takes at most microseconds to parse this. There are tons and tons of inefficiencies in both the design and current implementations, and that's fine because it's not meant to be particularly efficient.

> @tomaka all things considered, what do you reckon? do you prefer the break here or do you reckon https://github.com/paritytech/json-rpc-interface-spec/pull/66#discussion_r1270593268 is as clean of an API in the end?

> Even what's suggested in https://github.com/paritytech/json-rpc-interface-spec/pull/66#discussion_r1270593268 isn't really playing by the rules of the convention we've been discussing. When Pierre mentions "maybe some kind of identifier for the operation", what he's actually talking about is in fact a subscriptionId.

Indeed, what I suggest in that comment is to modify the problematic functions so that JSON-RPC library wouldn't implement as subscriptions but as simple request-responses. What the functions would return would still be exactly the same thing from a logical point of view as a "subscription ID", except in disguise.

It makes things a bit weirder, but it's solution designed to bypass this very specific problem.

josepot: Just to clarify, I'm struggling to see how Pierre's suggestion in [this comment](https://github.com/paritytech/json-rpc-interface-spec/pull/66#discussion_r1270593268) directly tackles the initial problem. The thing is, we'd still end up with a `subscriptionId` in the mix, which wouldn't be the only value inside the `result` property 🤷. That said, I totally see the value of Pierre's proposal in other ways - for instance, having a shared `stop` event seems like a great idea.

So, while I may not understand how this change makes our spec line up with the convention in question, or even if it brings a significant difference to the table, I'm definitely not against it. In fact, I quite like it (as long as we can still unsubscribe from those "nested subscriptions", ofc). If this change ends up serving multiple purposes, then that's a win-win, and I'm all for it.

jsdw: Thankyou both for your replies and patience with me!

I took the weekend to think about it, and I have definitely come around to this idea of removing the notion of "subscriptions" from the RPC client, leaving it to: 

- Handle the usual request/repsonse thing
- Allow the user to subscribe to notificaitons from the server and decide when/which notificaitons from the server to then pay attention to.

It'll then be up to the user to decide whether a given response means "ooh there may be notifications of interest to listen for now that will have this ID I can look out for" or whatever else. I think that's quite elegant, really :)

I have an API in mind for this that I'll have a play with.

That all said, of course if there is an equally clean way to express the APIs we want that we're all happy with which happens to also allow the current "subscription-aware" client/server to keep working as is for now, then I'd still be in favour of this just because it'd happen to avoid a bunch of work too and get this stuff implemented faster, but yup I'm good either way.

tomaka:
I've resolved the big conversation and pushed a bunch of commits for the second attempt.

src/api/chainHead_unstable_follow.md
```diff
@@ -129,6 +129,148 @@ All items in `finalizedBlockHashes` and `prunedBlockHashes` are guaranteed to ha
...
+
+In the situation where the `type` parameter was `"closest-descendant-merkle-value"`, the fact that `key` is equal to a `key` that was provided as parameter is necessary in order to avoid ambiguities when multiple `items` of type `"closest-descendant-merkle-value"` were requested.
+
+The `value` field is set if this item corresponds to one of the requested items whose `type` was `"value"` or `"descendants-values"`. The `value` field is a string containing the hexadecimal-encoded value of the storage entry.
+
+The `hash` field is set if this item corresponds to one of the requested items whose `type` was `"hash"` or `"descendants-hashes"`. The `hash` field is a string containing the hexadecimal-encoded hash of the storage entry.
+
+The `closest-descendant-merkle-value` field is set if this item corresponds to one of the requested items whose `type` was `"closest-descendant-merkle-value"`. The trie node whose Merkle value is indicated in `closest-descendant-merkle-value` is not indicated, as determining the key of this node might incur an overhead for the JSON-RPC server.
+
+### operation-waiting-for-continue
```
josepot, line 203: For consistency sake and for clarity, shouldn't this be named `operation-storage-waiting-for-continue`? 🤔

tomaka: I've omitted the `storage` because it's starting to be a bit long, and because in principle all operations could generate this event. Of course then I guess `chainHead_storageContinue` should be just `chainHead_continue`.

tomaka: Pushed more changes.

josepot: > and because in theory all operations could generate this eventt. Only `storage` can right now, but in version 2 maybe other operations could as well.

Ok, that makes sense.

> Of course then I guess `chainHead_storageContinue` should be just `chainHead_continue`.

Yes, please!

josepot (dismissed):
LGTM! Thanks @tomaka !

src/api/chainHead_unstable_follow.md
```diff
@@ -129,6 +129,148 @@ All items in `finalizedBlockHashes` and `prunedBlockHashes` are guaranteed to ha
 
 The current best block, in other words the last block reported through a `bestBlockChanged` event, is guaranteed to either be the last item in `finalizedBlockHashes`, or to not be present in either `finalizedBlockHashes` or `prunedBlockHashes`.
 
+### operation-body-done
+
+```json
+{
+    "event": "done",
```
lexnv, line 136: nit: Should the event here be `operation-body-done` to keep in sync with the explination from below and with `"event": "operation-storage-items"`?

josepot (approved):

jsdw:
Generally looks good to me, and I appreciate the compromise on subscription bits which will give us time to re-think the JSON-RPC bits without blocking this stuff being implemented :)

Just a couple of things I spotted:

- Should `archive_unstable_storageContinue` now be `archive_unstable_continue`?
- Related to a different PR but since I noticed it now: it looks like `archive_unstable_storage` is now out of sync with the `chainHead` version; should it be updated to be in sync?
- Now that methods send notifications via the `chainHead_follow` subscription, I guess a downside is that if the follow subscription ends with "stop" (eg becasue it can't keep up), I'll also no longer get back the block body or whatever that I had asked for I think?

tomaka:
> Should archive_unstable_storageContinue now be archive_unstable_continue?
> Related to a different PR but since I noticed it now: it looks like archive_unstable_storage is now out of sync with the chainHead version re how it works; should it be updated to be in sync?

The plan right now is work on and stabilize `chainHead` ASAP.

The `chainHead` functions fix the legacy JSON-RPC API in the sense that you can't properly do with the legacy API what you can do with `chainHead`.

`archive` however is just general clean ups, and nothing critical. If you're interested in doing `archive` calls you can use the legacy JSON-RPC API and it works.

For this reason I'm leaving it lagging behind a bit right now.

> Now that methods send notifications via the chainHead_follow subscription, I guess a downside is that if the follow subscription ends with "stop" (eg becasue it can't keep up), I'll also no longer get back the block body or whatever that I had asked for I think?

This is something I've taken into realized and I don't think it changes anything.

From the point of view of the JSON-RPC client, at any point in time a body/call/storage request can return `inaccessible` or `disjoint`. While the spec (before this PR) says that an operation continues even if `stop` is generated, a client actually has no guarantee that the operation will actually finish.

From the point of view of a light client server, the way it's implemented right now is: when body/call/storage is called, all the information necessary for the request is copied from the block and put in some local variables. Thanks to this, the request can continue even if the `chainHead_follow` dies.
Because right now a body/call/storage request can continue even when the `follow` dies, that's pretty much the only possible implementation strategy. You can't really do it differently than copying the information needed.
After this change, however, the implementation can be different and slightly more light. I don't think that in smoldot I will bother improving the implementation, but at least it could be possible.

jsdw (approved):
Thanks, that all makes sense to me; happy to see this merge!

tomaka:
> This effectively means that all operations are being disjoined, right?

Yes

PR #77, opened 2023-08-09
Rename the existing `transaction` to `transactionWatch` and add a new different `transaction` namespace

tomaka:
Close https://github.com/paritytech/json-rpc-interface-spec/pull/76

As mentioned in the title, the current `transaction` namespace is now `transactionWatch`, and a new `transaction` namespace is added, allowing just broadcasting transactions.

The new `transaction_unstable_broadcast` function is very dumb: it just broadcasts the transaction.
Before calling it, it is important for the JSON-RPC client to validate the transaction (using `chainHead_call`) and check if the `propagate` field returned by the validation is `true`. The JSON-RPC server intentionally does *not* do either of these things.
It is then the responsibility of the JSON-RPC client to download the bodies of blocks and check if the submitted transaction is there.

One drawback is that these new functions might be difficult to implement in Substrate.

cc @josepot @lexnv @jsdw


tomaka:
This function could presently be abused by malicious JSON-RPC clients to make the JSON-RPC server broadcast invalid transactions, leading to the JSON-RPC server losing reputation. A Substrate issue should be opened to remove this reputation loss.

tomaka:
I'm conflicted about whether the JSON-RPC server should perform a transaction validation.

Currently, a JSON-RPC client could abuse this function to make a JSON-RPC server broadcast invalid transactions. This will lead to the server losing reputation.

It could be a good idea to force the server to validate the transaction against its current best block (intentionally not saying which one) before broadcasting, but in that case the question is what to do if the transaction is invalid. It feels a bit stupid to just do nothing, but on the other hand reporting the problem to the JSON-RPC client leads to more complications.

jsdw:
I guess the point of this method is just so that if the client is already downloading and checking block bodies anyway, it can avoid calling `transactionWatch_submitAndWatch` and just rely on this?

(In Subxt's case, since it can talk to RPC nodes too, it's probably always preferable to call `transactionWatch_submitAndWatch` to avoid needing to download block bodies where possible, I think, so maybe this is more aimed at CAPI/light-client-_only_ APIs?).

Forcing validation and avoiding the client needing to do that step (and being able to broadcast invalid transactions etc) makes sense to me. What are the complications of returning an error if it is/becomes invalid?

josepot:
> Before calling it, it is important for the JSON-RPC client to validate the transaction (using `chainHead_call`) and check if the `propagate` field returned by the validation is `true`.

That's fine because we were already planing to use `chainHead_call` to call `system_dryRun` to check for possible errors before submitting the transaction.

> I guess the point of this method is just so that if the client is already downloading and checking block bodies anyway, it can avoid calling `transactionWatch_submitAndWatch` and just rely on this?
> 
> (In Subxt's case, since it can talk to RPC nodes too, it's probably always preferable to call `transactionWatch_submitAndWatch` to avoid needing to download block bodies where possible, I think, so maybe this is more aimed at CAPI/light-client-_only_ APIs?).
> 
> Forcing validation and avoiding the client needing to do that step (and being able to broadcast invalid transactions etc) makes sense to me. What are the complications of returning an error if it is/becomes invalid?

I'm of the opinion that for any library that wants to be light-client **first** (not light client **only**, but light-client first *cough* 🙂), the best approach would be to use the body of the finalized blocks as the only reliable source of information for knowing exactly when a transaction is in a finalized block. On top of that, it would also be possible to use the `transactionWatch_submitAndWatch` events as a "hint" in case that you get lucky and you receive an event **before** which would allow you to skip that check. That being said, I'm a bit on the fence on whether it would pay off to add the complexity of leveraging those "hints". I will probably try both approaches, though.

josepot (dismissed):
Awesome! Thanks @tomaka !

tomaka:
I've opened https://github.com/polkadot-fellows/RFCs/issues/19 for reference

josepot:
> The JSON-RPC server might check whether the transaction is valid before broadcasting it. If it does so and if the transaction is invalid, the server should silently do nothing and the JSON-RPC client is not informed of the problem. Invalid transactions should still count towards the limit to the number of simultaneously broadcasted transactions.

This is problematic.

I understand why you added this, but consider the following scenario: the JSON-RPC client is well informed about the latest `finalizedBlock`, it has validated the transaction against it and it has immediately broadcasted this transaction. It is possible that the JSON-RPC server is lagging behind a bit, right? So, in that case it would validate the transaction against the previous block, then the validation would fail, it wouldn't be broadcasted and the JSON-RPC client -despite having done everything correctly- would never know that's what happened.

I propose that if the server decides to validate and the validation fails, that we use the `operationId` for communicating that to the client via a notification, and that this notification includes in its payload the block that was used for performing the validation. Either that, or the server shouldn't be allowed to validate the transaction.

josepot:
@tomaka does my [previous comment](https://github.com/paritytech/json-rpc-interface-spec/pull/76#issuecomment-1672116779) make sense to you? 🙏 Also, I apologize for the number of edits 😬.

A possible alternative to the conditional event that I'm suggesting could be a non-conditional `broadcasted` event that should be emitted if the server has actually started broadcasting the event. I mean, I prefer the conditional event TBH, but a `broadcasted` event would also work.

jsdw:
I guess I still don't don't really see the use cases for this call over `transactionWatch_submitAndWatch`.

- `transactionWatch_submitAndWatch` will tell me if the transaction is broadcasted in case I'm interested in just that (in that case I'd stop paying attention after I saw that)
- `transactionWatch_submitAndWatch` will give me a clear end state that I know I will no longer need to do anything after (invalid/dropped/error/finalized). Whether I choose to scan block bodies myself or not, that's really useful to give me a better chance of not just being left scanning bodies for ages when the thing was invalid a while ago.
- If `transaction_unstable_broadcast` were only to "optionally" return an invalid error, that still means that it may not, which feels like a weaker promise than the other.

So then, the only use case I could see for this RPC call is as a "fire and forget" type thing. For me, this isn't useful, but perhaps for others it is. Am I missing something (quite probably this is the case right now :D)?

josepot:
> I guess I still don't don't really see the use cases for this call over `transactionWatch_submitAndWatch`.

Let me provide some insight.

**1. Significance of Latest Finalized Block:**  
To create a valid transaction, it's imperative to be informed of the latest finalized block. If you're crafting a transaction, you'll either leverage the `chainhead` group of functions or a non-spec compliant method to obtain the current finalized block.

**2. Dependency between Transaction and Chainhead Functions:**  
If a server can validate a transaction, it logically follows that the server can support the `chainhead` functions. The integral relationship between these two function sets shouldn't be overlooked. In essence, if a server offers transactional functionalities, it should also cater to `chainhead` functionalities, ensuring you construct transactions based on the same block against which they're validated.

**3. Argument against Segregation:**  
One might argue that separating the `transaction` functions from `chainhead` allows servers flexibility (some servers may opt-out from the `transaction` group). But if I had a say, this doesn't offer enough merit for keeping them distinct.

**4. Use Cases Illustration:**  
There's a spectrum of consumers, and understanding their needs underscores the importance of the suggested call:

- **The Casual User:** Some primarily aim to initiate transactions and ensure their inclusion in the chain, without an immediate need for the details. They’d immediately unpin every single block, while keeping track of the latest finalized block and would benefit from a method that abstracts away the intricacies. For them, `transactionWatch_submitAndWatch` is perfect.
  
- **The Detail-Oriented User:** On the contrary, some users must know immediately if their transaction is in the finalized (or best) block. They download block bodies, wanting to pinpoint their transactions instantly. For them, a low-level API makes more sense to eliminate redundancies between client and server actions.

Subxt, for instance, might not fit snugly into these categories, and that's alright. But CAPI aligns with the latter. If this PR doesn't get merged, we'll resort to using `transaction_submitAndWatch` and we will ignore the events that we don't need. However, we'd rather utilize the more streamlined API which precisely meets our requirements.

**Conclusion:**  
Reflecting on the arguments, I advocate for the integration of the `transaction` group into `chainhead`. It appears to be a solution that addresses the outlined concerns, as it would be possible to provide the necessary guarantees that are desirable when listening to the `transaction_submitAndWatch` events.

josepot:
As @tomaka pointed out on element, we should first have a proper discussion about this topic. I opened #79  as an attempt to have a structured discussion around this.

PR #80, opened 2023-08-17
transaction: add `searched` event

### LINKED ISSUE #79: Coupling between `transaction` and `chainhead` functions
### **Problem Statement**
Currently, the only method to send a transaction is through [`transaction_unstable_submitAndWatch`](https://paritytech.github.io/json-rpc-interface-spec/api/transaction_unstable_submitAndWatch.html#transaction_unstable_submitandwatch). This method undertakes several tasks:

- Validates the transaction against the latest finalized block.
- Broadcasts the transaction.
- Informs the consumer about the transaction's presence in the best and/or finalized block.

Despite its comprehensive functionalities, a significant challenge arises due to its interactions with another function: [`chaihead_unstable_follow`](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_unstable_follow.html#chainhead_unstable_follow). This function, belonging to a different group, provides crucial information about the latest finalized block, which is indispensable for creating valid transactions.

However, the inherent separation of these function groups and the lack of synchronization mechanisms pose risks:

1. **Lack of Synchronization**: If the `transaction` function lags behind the `chainhead`, it could result in users being unable to send transactions altogether.

2. **Delayed Events**: Events from the `transaction` function might be delayed or misaligned with the actual block status, rendering them ineffective for many applications. In such cases, users are forced to fetch block bodies manually and verify transaction presence, adding unnecessary complexity and duplicating efforts.

In essence, while the `chainhead` group of functions is properly isolated from other functionalities, the same can't be said for the `transaction` group. Its dependencies on the `chainhead` group bring up synchronization and information consistency concerns, which are pivotal for seamless transaction handling.

### **Proposed Solutions and Concerns**

**1. Tight Coupling of Function Groups**
- **Description**: Assume that the `transaction` group of functions is intrinsically linked to the `chainhead` group. Hence, if `transaction` is available, so is `chainhead`. This means that `transaction` would internally use `chainhead`.
- **Advantages**: Straightforward approach; ensures that users would always be able to send transactions and also ordered message arrival.
- **Drawbacks**: Challenges the very essence of function separation for load-balancers; cannot guarantee the presence of both function groups as per the spec.

**2. Decoupling `transaction` from `chainhead`**
- **Description**: The only role for the `transaction` group would be broadcasting a transaction. This would shift validation and block body transaction tracking responsibilities to the client.
- **Advantages**: Simplicity; forward compatibility due to its low-level nature.
- **Drawbacks**: Heavier responsibility on the client side; potential performance concerns; the API becomes challenging for users less acquainted with `chainhead`.

**3. Merging Function Groups**
- **Description**: Combine the two groups of functions (`transaction` and `chainhead`) into a single group, indicating their close interdependency.
- **Advantages**: Fewer events and edge cases.
- **Drawbacks**: Nodes may want to avoid transaction support. Allowing only a subset of functions in a group is currently not possible and also undesirable.

**4. Superset Approach with `transaction`**
- **Description**: Allow the `transaction` group of functions to encompass the responsibilities of the `chainhead` group, effectively making `transaction` a supergroup. Same solution as before, but leaving `chainhead` untouched.
- **Advantages**: Workaround to the function subset challenge.
- **Drawbacks**: Duplication.

I understand that these solutions aren't exhaustive and hope they provide insight into the possible ways forward. As a spec consumer, I find it crucial not only to highlight issues but also to offer potential resolutions. Eager to hear your insights on this.

josepot:
Close #79 

cc: @tomaka

### DIFF
--- a/src/api/transaction_unstable_submitAndWatch.md
+++ b/src/api/transaction_unstable_submitAndWatch.md
@@ -62,6 +62,34 @@ The JSON-RPC server doesn't (and can't) offer any guarantee that these peers hav
 
 If multiple `broadcasted` events happen in a row, the JSON-RPC server is allowed to skip all but the last.
 
+### `searched`
+
+The `searched` event signals that the JSON-RPC server has sought the transaction within a set of blocks.
+
+```json
+{
+    "event": "searched",
+    "blocks": [
+        {
+            "hash": "..."
+        },
+        {
+            "hash": "...",
+            "index": "..."
+        }
+    ]
+}
+```
+
+- `blocks`: An array of blocks in which the server has performed the search.
+  - `hash`: A string denoting the hexadecimal-encoded hash of the block's header.
+  - `index` (optional): A string-formatted integer representing the 0-based index of the transaction within the block. It's present only if the transaction exists in the block.
+
+**Notes:**
+
+- The JSON-RPC server does not guarantee the sequence in which it will scan the blocks.
+- While the server aims to dispatch the `searched` event as soon as possible, it doesn't ensure its delivery preceding any other event.
+
 ### bestChainBlockIncluded
 
 ```json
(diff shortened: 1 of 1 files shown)


josepot:
I went with an Array for the payload because I thought it might be easier for the server. But honestly, from the user's side, it doesn't make a huge difference to me. Whether we get multiple events one by one or a single event with a list of "searched" blocks, it's all good. If there's a way that works better for the server or makes things smoother, let's roll with that. Thoughts?

tomaka:
I haven't given a lot of thoughts to this yet, but two remarks:

- I'd be strongly in favor of merging this somehow with the other events. It feels very sloppy to have this completely separate even though the logic of this event directly interacts with the logic of other events.
- The server must be able to drop a transaction if the client doesn't pull events quickly enough. This is actually mildly annoying given that right now the server can only drop a transaction for chain-related reasons and not JSON-RPC-client-related reasons.
- It should be mentioned that these blocks might not overlap in any way with the blocks of `chainHead`, in other words that there's zero guarantee.

Also, this needs more modifications than just adding the event. There are multiple places in the spec that describe how transactions watching works, including the DoS resilience chapter.

josepot:
**Regarding merging the event with others:** 
I initially separated the event, thinking it'd cleanly add to the existing structure without messing with the current flow. But I get where you're coming from, and I see the appeal in consolidating the logic.

**On the server dropping transactions:** 
I'm trying to grasp how this new event could cause delays in pulling events. From my client-side perspective, I don't see it becoming an issue. But I can imagine situations where it might be problematic for clients on constrained devices. If that's a concern, maybe we could introduce an optional parameter in [`transaction_unstable_submitAndWatch`](https://paritytech.github.io/json-rpc-interface-spec/api/transaction_unstable_submitAndWatch.html#transaction_unstable_submitandwatch) like `withSearchedEvents`, defaulting to `false`. That way, only interested consumers would receive these events.

**Regarding the non-overlapping blocks with `chainHead`:** 
Totally missed that. Thanks for pointing it out!

**About the larger modifications:** 
Honestly, I was under the impression that integrating the event this way would simplify things, but I now realize the complexity. 

To be frank, I might not be the best person to take this forward. I just gave this PR a shot because I was [asked to](https://github.com/paritytech/json-rpc-interface-spec/issues/79#issuecomment-1681701271). Hopefully, this PR serves as a starting point for what consumers like me are looking for.

Given the situation, I think I'll close this PR and hand the reins over to @tomaka. If there's any chance that the `withSearchedEvents` parameter idea could revive this PR, do let me know, and I'll make the necessary updates. Otherwise, I'll be on standby for your PR, @tomaka. And no pressure to respond to all this – just putting my thoughts out there! 😊

PR #84, opened 2023-08-29
Use integer, not string for block index in `chainHead_follow` events

### LINKED ISSUE #83: `transaction_submitAndWatch`: consider returning "index" as integer, not string
Just a small thing for consideration; events like "finalized" look like:

```
{
    "event": "finalized",
    "block": {
        "hash": "0x...",
        "index": "2"
    }
}
``` 

The "index" is a string, not a number. Given that we never expect indexes to be particularly huge (certainly not outside what JS can interpret as a number), should we just return the integer instead?

Consistency is a reason for doing so, eg with this event on the same subscription, `numPeers` is an integer, not a string:

```
{
    "event": "broadcasted",
    "numPeers": ...
}
```

What's your preference?

jsdw:
Closes #83

/cc @tomaka @josepot @lexnv

### DIFF
--- a/src/api/transaction_unstable_submitAndWatch.md
+++ b/src/api/transaction_unstable_submitAndWatch.md
@@ -69,7 +69,7 @@ If multiple `broadcasted` events happen in a row, the JSON-RPC server is allowed
     "event": "bestChainBlockIncluded",
     "block": {
         "hash": "...",
-        "index": "..."
+        "index": ...
     }
 }
 ```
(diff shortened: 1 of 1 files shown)


lexnv (approved):

josepot (approved):
As I already mentioned before: this change makes sense to me. However, I would like to make sure that @tomaka is also on board with this.

tomaka (approved):
As you said, the API should be consistent. The `archive` API also uses hexadecimal strings to represent block heights. However, given that the maximum "safe" number in JavaScript is 2^53 – 1, it's safe to adjust the `archive` API as well.

jsdw:
I'll merge this since @lexnv's PR changes block heighs to integers :)

PR #90, opened 2023-09-11
archive/storage: Use a similar API with chainHead

lexnv:
The following changes are done to the archive methods:
 - Use integers for describing block heights instead of strings containing integers
 - Provide a similar API for `archive_storage` as for the `chainHead_storage`
 - Use camel case for event description 
 
### DIFF
--- a/src/api/archive_unstable_body.md
+++ b/src/api/archive_unstable_body.md
@@ -2,7 +2,7 @@
 
 **Parameters**:
 
-- `hash`: String containing the hexadecimal-encoded hash of the header of the block whose body must be retrieved.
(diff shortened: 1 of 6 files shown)


tomaka:
This should IMO be multiple different PRs. I don't see why they're bundled in one.

lexnv:
That's a good point! Thanks!

Closing this in favor of:
- https://github.com/paritytech/json-rpc-interface-spec/pull/92 
- https://github.com/paritytech/json-rpc-interface-spec/pull/93
- https://github.com/paritytech/json-rpc-interface-spec/pull/94
- https://github.com/paritytech/json-rpc-interface-spec/pull/95

PR #94, opened 2023-09-12
archive/storage: Backport API from chainHead

lexnv:
This PR backports the `chainHead_storage` method to `archive_storage` to bring it up to date.

The main changes include:
- support for batch requests via the introduced `items`
- removal of `includeDescendants` bool parameter in favor of `type` of the query
- support for fetching value, hash, closest merkle value, as well for iterating via descedantsValues and descendantsHashes

Differences between `chainHead_storage` and `archive_storage` include:
- `archive_storage` is a method that provides the responses directly, without any concepts of operationIds
- introduced an optional `paginationStartKey` key for descendants queries to resume pagination

### DIFF
--- a/src/api/archive_unstable_storage.md
+++ b/src/api/archive_unstable_storage.md
@@ -46,7 +46,7 @@ The JSON objects in the `"result"` field can have one of the following formats b
 
 The JSON object corresponds to one of the requested items whose `type` was `"value"` or `"descendantsValues"`.
 
-If the `key` is not associated with a storage value in the trie, then no response is generated.
+If the `key` is not associated with a storage value in the trie, then no response is generated in the `"result"` vector for this item.
 
 Returned when the `type` of the query was `"value"`:
 
@@ -69,7 +69,7 @@ Returned when the `type` of the query was `"descendantsValues"`:
 
 The JSON object corresponds to one of the requested items whose `type` was `"hash"` or `"descendantsHashes"`.
 
-If the `key` is not associated with a storage value in the trie, then no response is generated.
+If the `key` is not associated with a storage value in the trie, then no response is generated in the `"result"` vector for this item.
 
 Returned when the `type` of the query was `"hash"`:
 
@@ -93,7 +93,7 @@ Returned when the `type` of the query was `"descendantsHashes"`:
 
 The JSON object corresponds to one of the requested items whose `type` was `"closestDescendantMerkleValue"`.
 
-If the `key` doesn't exist in the trie, then the Merkle value of the closest descendant of `key` (including branch nodes) is provided. If `key` doesn't have any descendant in the trie, then no response is generated.
+If the `key` doesn't exist in the trie, then the Merkle value of the closest descendant of `key` (including branch nodes) is provided. If `key` doesn't have any descendant in the trie, then no response is generated in the `"result"` vector for this item.
 
 - `key` is guaranteed to be equal to one of the `key`s provided.
 - `closestDescendantMerkleValue` is the closest trie Merkle value of the `key`.


tomaka:
My opinion of `archive_storage` is actually to have it more similar to the legacy JSON-RPC API, and make it return just the list of values in its return value. Same for all the `archive`-prefixed functions in general.

The reasoning is that, compared to the `chainHead` functions, there's no "real time" component to `archive` functions.

lexnv:
I've updated the archive/storage to be a plain method, while keeping parity with the chainHead/storage interface.

One side effect of this being a method is the introduction of the `paginationStartKey` to avoid repeating a result. Would love to get your thoughts on this 🙏

src/api/archive_unstable_storage.md
```diff
@@ -3,91 +3,129 @@
...
-**Return value**: String containing an opaque value representing the operation, or `null` if no block with that `hash` exists.
+Each element in `items` must be an object containing the following fields:
+
+- `key`: String containing the hexadecimal-encoded key to fetch in the storage.
+- `type`: String equal to one of: `value`, `hash`, `closestDescendantMerkleValue`, `descendantsValues`, `descendantsHashes`.
+- `paginationStartKey`: This parameter is optional and should be a string containing the hexadecimal-encoded key from which the storage iteration should resume. This parameter is only valid in the context of `descendantsValues` and `descendantsHashes`.
+
+**Note**:
 
 The JSON-RPC server must obtain the value of the entry with the given `key` from the storage, either from the main trie of from `childTrie`. If `includeDescendants` is `true`, then the values of all the descendants must be obtained as well.
```
jsdw, line 17: Nit: maybe do as `chainHead_storage` does and prepend "For each item in `items`"

src/api/archive_unstable_storage.md
```diff
@@ -3,91 +3,129 @@
...
 
 If the block was previously returned by `archive_unstable_hashByHeight` at a height inferior or equal to the current finalized block height (as indicated by `archive_unstable_finalizedHeight`), then calling this method multiple times is guaranteed to always return non-null and always the same results.
```
jsdw, line 19: I'd omit the "If the block was previously returned" bit here; the hash may have come from that call or it may be known from elsewhere; it's not super relevant. So I'd focus on the relevant bit and write something like:

If the height of the block `hash` provided is less than or equal to the current finalized block height (which can be obtained via `archive_unstable_finalizedHeight`), then calling this method with the same parameters will always return the same response.

If the height of the block `hash` provided is greater than the current finalized block height, then the block might be pruned at any time and calling this method may return `null`.

src/api/archive_unstable_storage.md
```diff
@@ -3,91 +3,129 @@
...
 
-Yields an item that was found in the storage.
+The JSON object corresponds to one of the requested items whose `type` was `"value"` or `"descendantsValues"`
+
+Returned when the `type` of the query was `"value"`:
 
-The `key` field is a string containing the hexadecimal-encoded key of the value that was found.
-If the `includeDescendants` parameter was `true`, this `key` is guaranteed to start with the `key` provided as parameter.
-If the `includeDescendants` parameter was `false`, then it is also guaranteed to be equal to the `key` provided as parameter.
+- `key` is guaranteed to be equal to one of the `key`s provided as parameter
```
jsdw, line 60:
```suggestion
- `key` is guaranteed to be equal to one of the `key`s provided as a parameter
```

src/api/archive_unstable_storage.md
```diff
@@ -3,91 +3,129 @@
...
+- `value` is a string containing the hexadecimal-encoded value of the storage entry.
 
-The `value` field is a string containing the hexadecimal-encoded value of the storage item.
+Returned when the `type` of the query was `"descendantsValues"`:
 
-### waiting-for-continue
+- `key` is guaranteed to start with one of the `key`s provided as parameter
```
jsdw, line 65:
```suggestion
- `key` is guaranteed to start with one of the `key`s provided as a parameter
```

src/api/archive_unstable_storage.md
jsdw: (and below in a few places)

jsdw: Also nit: some of these end with a `.` and others don't :)

src/api/archive_unstable_storage.md
```diff
@@ -3,91 +3,129 @@
...
+The trie node whose Merkle value is indicated in `closestDescendantMerkleValue` is not indicated, as determining the key of this node might incur an overhead for the JSON-RPC server. The Merkle value is equal to either the node value or the hash of the node value, as defined in the [Polkadot specification](https://spec.polkadot.network/chap-state#defn-merkle-value).
 
-```json
-{
-    "event": "stop"
-}
-```
+## Overview
+
+For each item in `items`, the JSON-RPC server must start obtaining the value of the entry with the given `key` from the storage, either from the main trie or from `childTrie`. If `type` is `descendantsValues` or `descendantsHashes`, then it must also obtain the values of all the descendants of the entry.
```
jsdw, line 110:
```suggestion
For each item in `items`, the JSON-RPC server must obtain the value of the entry with the given `key` from the storage, either from the main trie or from `childTrie`. If `type` is `descendantsValues` or `descendantsHashes`, then it must also obtain the values of all the descendants of the entry.
```

src/api/archive_unstable_storage.md
```diff
@@ -3,91 +3,129 @@
...
+
+For the purpose of storage requests, the trie root hash of the child tries of the storage can be found in the main trie at keys starting the bytes of the ASCII string `:child_storage:`. This behaviour is consistent with all the other storage-request-alike mechanisms of Polkadot and Substrate-based chains, such as host functions or libp2p network requests.
+
+This function should be used when the target block is older than the blocks reported by `chainHead_unstable_follow`.
+Use `chainHead_unstable_storage` if instead you want to retrieve the storage of a block obtained by the `chainHead_unstable_follow`.
+
+If the `type` of an item is `value`, and `key` is associated with a storage value in the trie, then the result will include an item that contains this storage value. If `key` is not associated with a storage value in the trie, then the result will not include such item.
```
jsdw, line 117: These "If the type..." paragraphs here and below feel like a copy of the description of the possible formats above; maybe they can be pruned, and any extra information they might contain be moved to above? Or vice versa.

src/api/archive_unstable_storage.md
lexnv: That makes sense, have removed the `as a param` and added missing dots thanks!

src/api/archive_unstable_storage.md
lexnv: Have moved parts of the overview above, thanks!

lexnv:
@tomaka would love to get your feedback again on this 🙏

src/api/archive_unstable_storage.md
```diff
@@ -2,92 +2,120 @@
 
 **Parameters**:
 
-- `hash`: String containing a hexadecimal-encoded hash of the header of the block whose storage to fetch.
-- `key`: String containing the hexadecimal-encoded key to fetch in the storage.
+- `hash`: String containing an hexadecimal-encoded hash of the header of the block whose storage to fetch.
```
jsdw, line 5:
```suggestion
- `hash`: String containing a hexadecimal-encoded hash of the header of the block whose storage to fetch.
```

src/api/archive_unstable_storage.md
```diff
@@ -3,91 +3,119 @@
...
-- `includeDescendants`: Boolean indicating whether the key-values of all the descendants of the `key` should be returned as well.
 
-**Return value**: String containing an opaque value representing the operation, or `null` if no block with that `hash` exists.
+Each element in `items` must be an object containing the following fields:
+
+- `key`: String containing the hexadecimal-encoded key to fetch in the storage.
+- `type`: String equal to one of: `value`, `hash`, `closestDescendantMerkleValue`, `descendantsValues`, `descendantsHashes`.
+- `paginationStartKey`: This parameter is optional and should be a string containing the hexadecimal-encoded key from which the storage iteration should resume. This parameter is only valid in the context of `descendantsValues` and `descendantsHashes`.
+
+**Note**:
```
jsdw, line 15: For consistency with chainHead_storage, I think that this Note section should be moved to be right above the Possible Errors section.

(Also, I'd want to see the return value near the input args rather than below a bunch of text :))

jsdw: (and maybe rename it to "## Overview" too)

src/api/archive_unstable_storage.md
```diff
@@ -3,91 +3,119 @@
...
+This function should be used when the target block is older than the blocks reported by `chainHead_unstable_follow`.
+Use `chainHead_unstable_storage` if instead you want to retrieve the storage of a block obtained by the `chainHead_unstable_follow`.
 
-If the block was previously returned by `archive_unstable_hashByHeight` at a height inferior or equal to the current finalized block height (as indicated by `archive_unstable_finalizedHeight`), then calling this method multiple times is guaranteed to always return non-null and always the same results.
+If `items` contains multiple identical or overlapping queries, the JSON-RPC server can choose whether to merge or not the items in the result. For example, if the request contains two items with the same key, one with `hash` and one with `value`, the JSON-RPC server can choose whether to generate two `item` objects, one with the value and one with the hash, or only a single `item` object with both `hash` and `value` set. The JSON-RPC server is encouraged to notify as soon as possible of the information at its disposal, without waiting for missing information.
```
jsdw, line 27:
```suggestion
If `items` contains multiple identical or overlapping queries, the JSON-RPC server can choose whether to merge or not the items in the result. For example, if the request contains two items with the same key, one with `hash` and one with `value`, the JSON-RPC server can choose whether to generate two `item` objects, one with the value and one with the hash, or only a single `item` object with both `hash` and `value` set.
```

(I think that line isnt relevant to this now, since we dont stream results)

jsdw (dismissed):
Looking good to me now, nice one!

tomaka:
I'm a bit skeptical, because I feel like this is designed without a target use case in mind.
Who for example would need the `closestDescendantMerkleValue` of an old block? I really don't see why you'd want that information.

I have the impression that scanning the archive requires functions such as obtaining the list of blocks where a certain change to a storage value happened, and that kind of things, and I think that the `archive`-prefixed functions in general should be re-designed from the grounds up.

jsdw:
> Who for example would need the closestDescendantMerkleValue of an old block? I really don't see why you'd want that information.

One thing I have in mind is that I can fall back to using these functions in subxt in case blocks are unpinned or whatever via the chainHead calls. So basically, if archive fns are available, people can "hold onto" blocks for much longer than if only chainHead is available.

Hard to see how often that'll end up being used, but there are things (eg network issues) that can cause recent blocks to be unpinned, so nice to have a fallback where applicable

tomaka:
> if archive fns are available, people can "hold onto" blocks for much longer than if only chainHead is available.

The entire point of splitting between `archive` and `chainHead` is to specialize the ways the chain is accessed: either you're watching the chain live, or you're watching a replay.
Trying to create an abstraction that works transparently over both is wrong.

If you want the guarantee to be able to read any block even in case of networking issues and such, then just use `archive` all the time and not `chainHead`.

jsdw:
> Trying to create an abstraction that works transparently over both is wrong.

I don't agree; I think it's very reasonable for Subxt to use chainHead to follow and work with current blocks (by far the primary way people will use it), and if we happen to be talking to a node providing archive methods then we can do a little better and still attempt to return valid details for blocks that are no longer pinned via those, if the user has held on to them for longer for some reason.

This also means that when the user asks for some details at any block hash, subxt will use a pinned block if one is available, or fall back to using an archive call if that is available, or fail if neither is possible. It makes the experience simple and does a "best effort" to retrieve the requested details. 

Subxt doesn't try to guarantee anything; instead it simply aims to keep the details for any block that the user is interested in available for as long as possible.

tomaka:
> It makes the experience simple and does a "best effort" to retrieve the requested details.

I really disagree with that.

If SubXT can "sometimes" return information about a block and "sometimes" can't, it doesn't make it simple. Users needs to know when what they're doing is going to work or not.

Telling the user that you're going to *try* give them some information but without any guarantee ultimately leads either to a bad user experience or to the user having to punch through the abstraction layers in order to understand how things work under the hood.

jsdw:
> If SubXT can "sometimes" return information about a block and "sometimes" can't, it doesn't make it simple. Users needs to know whether their code is going to work or not, and shrugging at them isn't a very satisfactory thing to do.

Subxt offers the best guarantees that I think it can here; it will keep blocks pinned for as long as it can while the user is interested in them, and so as long as you ask for information from a block within a reaosnable time frame, it will always successfully return it. The new API only guarantees that blocks will be pinned for some (currently unspecified and unknown) amount of time, and so if the user holds onto one for too long, I'll eventually have to say "no". What do you think would be a better approach here?

My thought then was that if the user _does_ hold onto a block reference for too long, and they are specifically connected to an archive node (presumably this would be a deliberate choice, especially as light client connections will eventually be the default) then they can, though exactly the same Subxt APIs, continue access information about older blocks too, in a seamless sort of way.

> It is clear and precise enough that users of the API don't have to look under the hood in order to understand whether their request is going to work.

A user has to be aware that a request will work for some period of time after a block hash is seen. Currently, they won't know how long that period is, but they will know when it's no longer possible because a "stop" event is emitted. Subxt offers the same promise; if you access details about some block within a reasonable time frame, the access will succeed. We don't currently allow the user to be notified when some given block ref is no longer pinned/accessible, but that's definitely something we can add.

> The entire point of splitting between archive and chainHead is to specialize the ways the chain is accessed: either you're watching the chain live, or you're watching a replay.

It's not clear to me at least what should be different between archive and chainhead method; do you have any ideas on how the archive methods should differ?

I'd likely offer a separate set of APIs for accessing archived block information in Subxt if the methods did differ in a meaningful way, but it would add complexity, and the user would then have to learn two sets of APIs (one for recent blocks and one for old blocks), so I do think that there would have to be at least a slightly compelling reason for any significant differences, but I'm open to all thoughts here!

jsdw:
> To put it otherwise, to me the solution is for example for archive nodes to allow an unlimited amount of pinned blocks, provided that these blocks are in the finalized canonical chain. In other words, the blocks that you would have been able to access through archive anyway can remain pinned.

That makes a lot of sense to me!

I still think that Subxt could provide a single interface to allow fetching of current or old blocks, but could imagine that a user might explicitly pick which of these they are doing and select the appropriate backend for the job (in other words; instruct Subxt to use the archive methods or to use the chainHead methods to drive things). I guess it depends on how different the `chainHead_` vs `archive_` interfaces ended up being to how I'd design this stuff. 

> For archive it's unfortunately way less clear to me how to design it. The users of this namespace would probably be people who use indexers for example, but I don't know precisely what these people want. For example https://github.com/paritytech/json-rpc-interface-spec/issues/108 is a legitimate wish, and I don't think I have enough knowledge to anticipate all wishes.

In the absence of more information, I think that it'd be good to merge this PR and start with the archive and chainHead methods being reasonably aligned (aside from the subscription vs request sort of approach, which makes sense to differ as this PR has done).

Then, we can iterate on the design from there. 

I can imagine that a bunch of things (like #108) may want to exist in chainHead and archive namespaces if possible (albeit with perhaps a slightly different approach to cater for each API; chainHead maybe giving a "live" diff each block for instance, vs archive just being given the block numbers to diff between up front).

lexnv:
> In the absence of more information, I think that it'd be good to merge this PR and start with the archive and chainHead methods being reasonably aligned ..
> Then, we can iterate on the design from there.

That sounds like a plan to me, and follow up with use cases for the archive class to make informed decisions about what developers need. @tomaka what do you think?

lexnv:
@tomaka @josepot would love to get your thoughts on this PR
 
As a first step having the `archive` slightly aligned with the `chainHead` would allow us to iterate a bit faster on the design; and would also unblock the substrate side implementation (https://github.com/paritytech/polkadot-sdk/pull/1846). Then we could gather more data about API usage and add/extend storage differences for example (https://github.com/paritytech/json-rpc-interface-spec/issues/108).

tomaka:
To me what you're suggesting is to change from a bad design that shouldn't be used to a different bad design that shouldn't be used. I don't have an opinion on this.

jsdw:
To me, this PR improves on the state of the `archive_` methods by making them more consistent with the `chainHead` ones. I think that this is a good place to aim for in the absense of any data supporting doing something different.

So let's merge this by the end of the week unless there are any remaining objections that we can solve, and when we get more data we can make the necessary changes.

tomaka:
Again I'm neutral here, but to me consistency between `archive` and `chainHead` is not an objective.

jsdw:
> consistency between archive and chainHead is not an objective. They're two completely different APIs for two completely different usages.

Agreed; not an objective, but a good default when we don't know better :)

src/api/archive_unstable_storage.md
```diff
@@ -3,91 +3,119 @@
...
-    "event": "item",
     "key": "0x0000000...",
     "value": "0x0000000...",
 }
 ```
 
-Yields an item that was found in the storage.
+The JSON object corresponds to one of the requested items whose `type` was `"value"` or `"descendantsValues"`.
+
+If the `key` is not associated with a storage value in the trie, then no response is generated.
```
niklasad1, line 49: I guess this is fine but would be easier for clients if the server would just send back `null` to indicate that key was not found.

For this API, a client probably needs to have a timeout for these request and after for instance 1 min the call is regarded as failed.

src/api/archive_unstable_storage.md
```diff
@@ -3,91 +3,119 @@
...
-    "event": "waiting-for-continue"
+    "key": "0x0000000...",
+    "hash": "0x0000000...",
 }
 ```
 
-The `waiting-for-continue` event is generated after at least one `"item"` event has been generated, and indicates that the JSON-RPC client must call `archive_unstable_storageContinue` before more events are generated.
+The JSON object corresponds to one of the requested items whose `type` was `"hash"` or `"descendantsHashes"`.
+
+If the `key` is not associated with a storage value in the trie, then no response is generated.
```
niklasad1, line 72: same here :point_up_2:

src/api/archive_unstable_storage.md
```diff
@@ -3,91 +3,119 @@
...
 
-This event only ever happens if the `includeDescendants` parameter was `true`.
+Returned when the `type` of the query was `"hash"`:
 
-While the JSON-RPC server is waiting for `archive_unstable_storageContinue` to be called, it can generate a `stop` event indicating that it can no longer proceed with that storage access.
+- `key` is guaranteed to be equal to one of the `key`s provided.
+- `hash` is a string containing the hexadecimal-encoded cryptographic hash of the storage entry.
```
niklasad1, line 77: picking hairs, `twox` is not a cryptographic hash function and it seems unnecessary to specify that here.

```suggestion
- `hash` is a string containing the hexadecimal-encoded hash of the storage entry.
```

src/api/archive_unstable_storage.md
```diff
@@ -3,91 +3,119 @@
...
+
+This function should be used when the target block is older than the blocks reported by `chainHead_unstable_follow`.
+Use `chainHead_unstable_storage` if instead you want to retrieve the storage of a block obtained by the `chainHead_unstable_follow`.
+
+If `items` contains multiple identical or overlapping queries, the JSON-RPC server can choose whether to merge or not the items in the result. For example, if the request contains two items with the same key, one with `hash` and one with `value`, the JSON-RPC server can choose whether to generate two `item` objects, one with the value and one with the hash, or only a single `item` object with both `hash` and `value` set.
 
-The `stop` event can be generated after a `waiting-for-continue` event in order to indicate that the JSON-RPC server can't continue. The JSON-RPC client should simply try again.
+It is allowed (but discouraged) for the JSON-RPC server to provide the same information multiple times in the result, for example providing the `value` field of the same `key` twice. Forcing the JSON-RPC server to de-duplicate items in the result might lead to unnecessary overhead.
```
niklasad1, line 117: In other words, the client is responsible to check the results and take care of duplicated items such as `value`, `hash`and so on?

My two cents is that the behavior would be specified here because we end up with a similar scenario as  "batch requests in JSON-RPC spec" where servers can decide whether to answer it in order or not but it may be more inefficient in some scenarios.

As client you don't know what the server will do and always have check it anyway to on the safe side, probably not that tricky in this case.

niklasad1:
> For archive it's unfortunately way less clear to me how to design it. The users of this namespace would probably be people who use indexers for example, but I don't know precisely what these people want. For example https://github.com/paritytech/json-rpc-interface-spec/issues/108 is a legitimate wish, and I don't think I have enough knowledge to anticipate all wishes.

> Furthermore there are many uncertainties w.r.t. how to manage the old chain. For example we might want to deprecate things in the future, which means maybe breaking old blocks. Is that acceptable? Is that something to bake in the API? I have no idea.

Yeah, I agree with :point_up: there lots of unknowns for sure  but on the hand we have to start somewhere to know what issues that comes up to "fetch old blocks" with this `archive API`.

It doesn't have to be perfect but as long we can get folks to move away from the current RPC API "to fetch old storage" and get feedback for users what doesn't work etc.. to iterate further on it.

niklasad1 (dismissed):

src/api/archive_unstable_storage.md
lexnv: Will modify the phrasing of the spec. What I meant by this, given archive/storage is a plain method, no response will be generated for this "key" in the "result" array. Users will still get back once all queries are processed the responses back

niklasad1 (approved):

src/api/archive_unstable_storage.md
niklasad1: Yeah, I agree. 

The client can just ensure that no duplicated items are request and don't have to worry about it.

PR #95, opened 2023-09-12
archive/body: Keep description consistent with chainHead

lexnv:
This PR uses the same description for the `hash` parameter as the one from the `chainHead`.

### DIFF
--- a/src/api/archive_unstable_body.md
+++ b/src/api/archive_unstable_body.md
@@ -2,7 +2,7 @@
 
 **Parameters**:
 
-- `hash`: String containing the hexadecimal-encoded hash of the header of the block whose body must be retrieved.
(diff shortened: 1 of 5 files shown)


tomaka (approved):

src/api/archive_unstable_body.md
```diff
@@ -2,7 +2,7 @@
 
 **Parameters**:
 
-- `hash`: String containing the hexadecimal-encoded hash of the header of the block whose body must be retrieved.
+- `hash`: String containing an hexadecimal-encoded hash of the header of the block whose body to fetch.
```
jsdw, line 5: Nit: Ideally I think it would read 

```suggestion
- `hash`: String containing a hexadecimal-encoded hash of the header of the block whose body will be retrieved.
```

But whatever it is, it's good to be consistent!

jsdw (dismissed):

PR #98, opened 2023-09-15
Use the same method name for notification results

lexnv:
Currently, there's a method suffixed with `Event` that generates JSON-RPC notifications (ie subscription results).
For example, the subscription events generated by calling `chainHead_unstable_follow` are provided under `"method": "chainHead_unstable_followEvent"`.

This PR aligns the method name of the subscription that generates notifications with the method called, keeping parity with Substrate (pub/sub model used).
Also, it is not possible to configure this in [jsonrpsee](https://github.com/paritytech/jsonrpsee/blob/96c035c56094606fe98b98f9a033cced32781d04/core/src/server/rpc_module.rs#L283-L286).

Would love to hear your thoughts 🙏 

cc @tomaka @josepot @jsdw @paritytech/subxt-team

### DIFF
--- a/src/api/archive_unstable_storage.md
+++ b/src/api/archive_unstable_storage.md
@@ -22,7 +22,7 @@ This function will later generate notifications in the following format:
 ```json
 {
     "jsonrpc": "2.0",
-    "method": "archive_unstable_storageEvent",
+    "method": "archive_unstable_storage",
     "params": {
         "subscription": "...",
         "result": ...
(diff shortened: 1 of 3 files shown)


tomaka:
> This PR aligns the method name of the subscription that generates notifications with the method called, keeping parity with Substrate (pub/sub model used).

I'm not sure I understand, because this is not the case for the legacy API.
For example, `state_subscribeStorage` generates events whose method is `state_storage`, `author_submitAndWatchExtrinsic` generates events whose method is `author_extrinsicUpdate`, and so on.

lexnv:
Ops my bad! Indeed, they could be mapped to a different name, will rename them in substrate! Thanks 👍

PR #111, opened 2023-11-13
chainHead/unpin: Explicit about unpin parameters

lexnv:
This PR renames the parameter of the unpin method from `hash` to `hash_or_hashes` to suggest to the users that it can be one of two different formats: plain hex or array of hexes.

Another approach to clarify this would be to change the `hash` parameter to `hashes` and accept only an array of hex-encoded hashes. 

// cc @paritytech/subxt-team @josepot @tomaka


niklasad1 (approved):

tomaka (approved):
🤷

tomaka:
Actually, it should have been `hashOrHashes`

PR #114, opened 2023-11-20
Add archive node description

IkerAlus:
The description for archive nodes was missing.

### DIFF
--- a/src/grouping-functions-and-node-capabilities.md
+++ b/src/grouping-functions-and-node-capabilities.md
@@ -8,7 +8,7 @@ The group a JSON-RPC function belongs to is indicated by the `prefix_` in its na
 
 This JSON-RPC interface is intended to be implemented on various types of nodes:
 
-- Full nodes, which hold in their database all headers from the genesis block to the most recent block, plus the bodies and storage of the most recent blocks (typically the latest 256).
+- Full nodes, which hold in their database all block headers and bodies, plus the storage of recent blocks.
 - Light clients, which only know the headers of recent blocks.
 - Archive nodes, which hold in their database all block headers, bodies, and storage of every past block.
 - Plain databases that aren't actually connected to the peer-to-peer network of the blockchain.
--- a/src/grouping-functions-and-node-capabilities.md
+++ b/src/grouping-functions-and-node-capabilities.md
@@ -10,7 +10,7 @@ This JSON-RPC interface is intended to be implemented on various types of nodes:
 
 - Full nodes, which hold in their database all block headers and bodies, plus the storage of recent blocks.
 - Light clients, which only know the headers of recent blocks.
-- Archive nodes, which hold.
+- Archive nodes, which hold in their database all block headers and bodies, plus the storage of every past block.
 - Plain databases that aren't actually connected to the peer-to-peer network of the blockchain.
 
 These various node implementations have different capabilities, and it is normal for some implementations to only support some functions and not others. It doesn't make sense for example for a light client to support a function that rotates keys.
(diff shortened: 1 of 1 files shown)


src/grouping-functions-and-node-capabilities.md
```diff
@@ -10,7 +10,7 @@ This JSON-RPC interface is intended to be implemented on various types of nodes:
 
 - Full nodes, which hold in their database all block headers and bodies, plus the storage of recent blocks.
 - Light clients, which only know the headers of recent blocks.
-- Archive nodes, which hold.
+- Archive nodes, which hold in their database all block headers and bodies, plus the storage of every past block.
```
josepot, line 13: Thanks for noticing this @IkerAlus !

However, I'm afraid that the explanation of "Full nodes" is also off. It should probably say something more or less like:

> - Full nodes, which hold in their database all from the genesis block to the most recent block, plus the bodies and storage of the most recent blocks (typically the latest 256).

Basically, I'm quite certain that it's not true that "full nodes" hold all the bodies from genesis. I'm pretty sure that they only store the headers, which is what it's needed to validate that the state is correct.

Then IMO the "Archive nodes" description should read something more or less like this:

> - Archive nodes, which hold in their database all block headers, bodies, and storage of every past block.

IkerAlus: yes, I think you are right respect to full nodes description.  

I will change it in this PR too. Thanks!

src/grouping-functions-and-node-capabilities.md
```diff
@@ -8,9 +8,9 @@ The group a JSON-RPC function belongs to is indicated by the `prefix_` in its na
 
 This JSON-RPC interface is intended to be implemented on various types of nodes:
 
-- Full nodes, which hold in their database all block headers and bodies, plus the storage of recent blocks.
+- Full nodes, which hold in their database all headers from the genesis block to the most recent block, plus the bodies and storage of the most recent blocks (typically the latest 256).
```
tomaka, line 11: Full nodes store the bodies of all blocks. The sentence was accurate, and the number of blocks that are kept is unstable and arbitrary and thus better left out.

```suggestion
- Full nodes, which hold in their database all block headers and bodies, plus the storage of recent blocks.
```

josepot: Oh wow! I didn't know that! If that's the case, then I suppose that [the substrate docs](https://docs.substrate.io/learn/networks-and-nodes/#full-nodes) should also be updated, because they clearly state that:

> By default, full nodes are configured to store only the most recent 256 blocks and to discard state older than that—with the exception of the genesis block—to prevent the full node from growing indefinitely and consuming all available disk space. You can configure the number of blocks a full node retains.

> Although older blocks are discarded, full nodes retain all of the [block headers](https://docs.substrate.io/reference/glossary/#header) from the genesis block to the most recent block to validate that the state is correct.

josepot: Actually, the next sentence already hints that full-nodes also store the bodies:
 
> Because the full node has access to all of the block headers, it can be used to rebuild the state of the entire blockchain by executing all of the blocks from the genesis block. Thus it requires much more computation to retrieve information about some previous state, and an archive should generally be used instead.

If they are able to rebuild the state, that also implies that they have the bodies (transactions) and not just the headers. So, yeah, I think that those docs are a bit misleading. Thanks @tomaka !

IkerAlus: I accepted @tomaka 's suggestion so that this PR is about adding the missing description only as originally created.

josepot: Wait! I have substrate developers telling me that:

> Block bodies should also be discarded depending on the --blocks-pruning setting of the node.

josepot: image

josepot: Ok, it's good to know that I wasn't the only one under that wrong impression! So, yeah, we should probably improve the substrate docs. 🙂

PR #117, opened 2023-11-27
chainHead: Define error codes returned by the server

lexnv:
This PR defined the error codes returned by the server:
- `-32800` two or more subscriptions requested for chainHead_follow
- `-32801` block hash not reported by chainHead_follow or block hash has been unpinned
- `-32802` chainHead_follow started with withRuntime == false
- `-32803` chainHead_follow did not generate an operationWaitingForContinue event

The following are errors defined in the [JSON-RPC spec](https://www.jsonrpc.org/specification#error_object):
- `-32602` The provided parameter isn't one of the expected values, has different format or is missing
- `-32603` Internal server error:
  - The server was unable to generate a valid subscription ID (https://github.com/paritytech/polkadot-sdk/blob/21f1811c6600d8a7fe043592ff34dcb79284d583/substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs#L198-L204)
  - The server cannot interpret the block: hardware error, ram/disk error, corrupt database, broken code logic (https://github.com/paritytech/polkadot-sdk/blob/21f1811c6600d8a7fe043592ff34dcb79284d583/substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs#L306-L309)

This is done for consistency reasons, since at the moment the `chainHead_follow` is the only method that has an associated return code, although the other methods could still generate JSON-RPC errors.
This ensures interoperability as both substrate and light-client implementations will return similar codes.

cc @tomaka @jsdw @josepot @bkchr

### DIFF
--- a/src/api/chainHead_unstable_body.md
+++ b/src/api/chainHead_unstable_body.md
@@ -51,5 +51,4 @@ This function should be seen as a complement to `chainHead_unstable_follow`, all
 
 - If the networking part of the behaviour fails, then a `{"event": "operationInaccessible"}` notification is generated (as explained above).
 - If the `followSubscription` is invalid or stale, then `"result": "limitReached"` is returned (as explained above).
-- A JSON-RPC error is generated if the block hash passed as parameter doesn't correspond to any block that has been reported by `chainHead_unstable_follow`.
-- A JSON-RPC error is generated if the `followSubscription` is valid but the block hash passed as parameter has already been unpinned.
+- A JSON-RPC error with error code `-32801` is generated if the block hash passed as parameter doesn't correspond to any block that has been reported by `chainHead_unstable_follow`, or the block hash has been unpinned.
--- a/src/api/chainHead_unstable_body.md
+++ b/src/api/chainHead_unstable_body.md
@@ -53,4 +53,4 @@ This function should be seen as a complement to `chainHead_unstable_follow`, all
 - If the `followSubscription` is invalid or stale, then `"result": "limitReached"` is returned (as explained above).
 - A JSON-RPC error with error code `-32801` is generated if the block hash passed as parameter doesn't correspond to any block that has been reported by `chainHead_unstable_follow`, or the block hash has been unpinned.
 - A JSON-RPC error with error code `-32602` is generated if one of the parameters doesn't correspond to the expected type (similarly to a missing parameter or an invalid parameter type).
-- A JSON-RPC error with error code `-32805` is generated if the JSON-RPC server cannot interpret the block (hardware issues, corrupted database, disk failure etc).
+- A JSON-RPC error with error code `-32603` is generated if the JSON-RPC server cannot interpret the block (hardware issues, corrupted database, disk failure etc).
--- a/src/api/chainHead_unstable_body.md
+++ b/src/api/chainHead_unstable_body.md
@@ -52,5 +52,5 @@ This function should be seen as a complement to `chainHead_unstable_follow`, all
 - If the networking part of the behaviour fails, then a `{"event": "operationInaccessible"}` notification is generated (as explained above).
 - If the `followSubscription` is invalid or stale, then `"result": "limitReached"` is returned (as explained above).
 - A JSON-RPC error with error code `-32801` is generated if the block hash passed as parameter doesn't correspond to any block that has been reported by `chainHead_unstable_follow`, or the block hash has been unpinned.
-- A JSON-RPC error with error code `-32804` is generated if one of the parameters doesn't correspond to the expected type (similarly to a missing parameter or an invalid parameter type).
+- A JSON-RPC error with error code `-32602` is generated if one of the parameters doesn't correspond to the expected type (similarly to a missing parameter or an invalid parameter type).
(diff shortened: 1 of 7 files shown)


src/api/chainHead_unstable_body.md
```diff
@@ -51,5 +51,6 @@ This function should be seen as a complement to `chainHead_unstable_follow`, all
 
 - If the networking part of the behaviour fails, then a `{"event": "operationInaccessible"}` notification is generated (as explained above).
 - If the `followSubscription` is invalid or stale, then `"result": "limitReached"` is returned (as explained above).
-- A JSON-RPC error is generated if the block hash passed as parameter doesn't correspond to any block that has been reported by `chainHead_unstable_follow`.
-- A JSON-RPC error is generated if the `followSubscription` is valid but the block hash passed as parameter has already been unpinned.
+- A JSON-RPC error with error code `-32801` is generated if the block hash passed as parameter doesn't correspond to any block that has been reported by `chainHead_unstable_follow`, or the block hash has been unpinned.
+- A JSON-RPC error with error code `-32804` is generated if one of the parameters doesn't correspond to the expected type (similarly to a missing parameter or an invalid parameter type).
+- A JSON-RPC error with error code `-32805` is generated if the JSON-RPC server cannot interpret the block (hardware issues, corrupted database, disk failure etc).
```
tomaka, line 56: That would be an internal server error, no?

src/api/chainHead_unstable_body.md
```diff
@@ -51,5 +51,6 @@ This function should be seen as a complement to `chainHead_unstable_follow`, all
 
 - If the networking part of the behaviour fails, then a `{"event": "operationInaccessible"}` notification is generated (as explained above).
 - If the `followSubscription` is invalid or stale, then `"result": "limitReached"` is returned (as explained above).
-- A JSON-RPC error is generated if the block hash passed as parameter doesn't correspond to any block that has been reported by `chainHead_unstable_follow`.
-- A JSON-RPC error is generated if the `followSubscription` is valid but the block hash passed as parameter has already been unpinned.
+- A JSON-RPC error with error code `-32801` is generated if the block hash passed as parameter doesn't correspond to any block that has been reported by `chainHead_unstable_follow`, or the block hash has been unpinned.
+- A JSON-RPC error with error code `-32804` is generated if one of the parameters doesn't correspond to the expected type (similarly to a missing parameter or an invalid parameter type).
```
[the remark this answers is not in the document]
lexnv: That's a good point, will update the return code to rely on the JSON-RPC spec (`-32602 	Invalid params`), thanks 🙏

src/api/chainHead_unstable_body.md
lexnv: Yep, I'll update this to `-32603 	Internal error`, thanks for the review 🙏

src/api/chainHead_unstable_call.md
```diff
@@ -61,7 +61,7 @@ This function should be seen as a complement to `chainHead_unstable_follow`, all
 - If the runtime call fails (e.g. because it triggers a panic in the runtime, running out of memory, etc., or if the runtime call takes too much time), then an `{"event": "operationError"}` notification is generated.
 - A JSON-RPC error with error code `-32801` is generated if the block hash passed as parameter doesn't correspond to any block that has been reported by `chainHead_unstable_follow`, or the block hash has been unpinned.
 - A JSON-RPC error with error code `-32802` is generated if the `followSubscription` corresponds to a follow where `withRuntime` was `̀false`.
```
tomaka, line 63:
```suggestion
- A JSON-RPC error with error code `-32802` is generated if the `followSubscription` corresponds to a follow where `withRuntime` was `false`.
```

tomaka (approved):

tomaka:
> The server was unable to generate a valid subscription ID (https://github.com/paritytech/polkadot-sdk/blob/21f1811c6600d8a7fe043592ff34dcb79284d583/substrate/client/rpc-spec-v2/src/chain_head/chain_head.rs#L198-L204)

Note that this shouldn't be a legitimate error. Subscription IDs can be strings, so there is an infinite number of possible IDs. Even if it wasn't infinite, the spec mentions that there's a maximum number of simultaneous `chainHead_follow` subscriptions.

src/api/chainHead_unstable_follow.md
```diff
@@ -400,3 +400,5 @@ The runtime is of type `invalid` if the JSON-RPC server considers the runtime as
 ## Possible errors
 
 - A JSON-RPC error with error code `-32800` can be generated if the JSON-RPC client has already opened 2 or more `chainHead_unstable_follow` subscriptions.
+- A JSON-RPC error with error code `-32602` is generated if one of the parameters doesn't correspond to the expected type (similarly to a missing parameter or an invalid parameter type).
+- A JSON-RPC error with error code `-32603` is generated if the JSON-RPC server was unable of generating a valid subscription ID.
```
tomaka, line 404:
```suggestion
```

tadeohepperle (approved):
:+1:

lexnv:
Thanks for the reviews 🙏

PR #121, opened 2024-01-05
docs: improve `operationInaccessible` event

josepot:
cc @tomaka

### DIFF
--- a/src/api/chainHead_unstable_follow.md
+++ b/src/api/chainHead_unstable_follow.md
@@ -245,7 +245,7 @@ No more event will be generated with this `operationId`.
 }
 ```
 
-`operationId` is a string returned by `chainHead_unstable_body`, `chainHead_unstable_call`, or `chainHead_unstable_storage`.
+`operationId` is, in the case of this event, a string returned only by `chainHead_unstable_storage`.
 
 The `operationInaccessible` event is produced if the JSON-RPC server was incapable of obtaining the storage items necessary for the given operation.
 


tomaka:
No, the sentence is correct, I don't see why you think that this event is specific to storage

josepot:
> No, the sentence is correct, I don't see why you think that this event is specific to storage

Because the next sentence states that:

> The `operationInaccessible` event is produced if the JSON-RPC server was incapable of obtaining the storage items necessary for the given operation.

josepot:
I'm guessing that the other operation that can also trigger this event must be [`chainHead_unstable_body`](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_unstable_body.html#chainhead_unstable_body), since it's probably just sugar for accessing the storage entry of `System.ExtrinsicData`? If that's the case, I'm starting to think that we should probably remove `chainHead_unstable_body` from the API.

tomaka:
You're overthinking it, and it's just the next sentence that must mention that it can happen if failing to retrieve the body.

> I'm guessing that the other operation that can also trigger this event must be [chainHead_unstable_body](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_unstable_body.html#chainhead_unstable_body), since it's probably just sugar for accessing the storage entry of System.ExtrinsicData? If that's the case, I'm starting to think that we should probably remove chainHead_unstable_body from the API.

It's not just sugar. If anything, it's `System.ExtrinsicData` that should be removed.

> Ahhh, wait, of course: [chainHead_unstable_call](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_unstable_call.html#chainhead_unstable_call) also requires from a storage access to :code, so I guess that if the storage read is failing, then it will also trigger this event... I see 🤔

`call` reads the storage from the network because doing a runtime call accesses the storage.

PR #124, opened 2024-01-18
Add `chainSpec_unstable_spec` method

lexnv:
This PR adds the `chainSpec_unstable_spec` method to fetch the chainSpec from a running node.

Substrate nodes expose currently `sync_state_genSyncSpec`. The new method unifies the format of substrate chain-spec needed for the light-clients to sync to the head of the chain.

There are several differences from the traditional chain-spec:
- the `Genesis` entry contains either a `Raw` entry or a `stateRootHash` entry
  - the `stateRootHash` represents the merkle value of the genesis. This is an optimisation used by smoldot to store smaller chainSpec files, and to allow light-clients to sync up faster (since a raw entry must be first converted to a `stateRootHash` entry).
- `protocolID` and `forkId` are placed under a common `networkProperties` object
- `para_id` and `relay_chain` are placed under a common `parachain` object; while using camelCase naming
- `telemetryEndpoints` contains objects instead of tuples to explicitly describe the `address` of the telemetry server and the `verbosity_level`
- a new `checkpoint` object is added
  - the `forkedBlocks` entry is renamed to `trustedBlocks` and moved under this object
    - `forkedBlocks` are now identified by a block number and block hash
  - the `badBlocks` are moved under this object
  

The `checkpoint` object could then be extended to include the syncState needed by lightclients: https://github.com/paritytech/substrate/issues/11184

Block numbers are represented as string unsigned integers, similar to the current chainSpec. I believe this may be an artifact of javaScript users.

// cc @jsdw @tomaka @bkchr @skunert

### DIFF
--- a/src/api/chainSpec_unstable_spec.md
+++ b/src/api/chainSpec_unstable_spec.md
@@ -14,6 +14,8 @@ The JSON object returned by this function has the following format:
 
     "id": "...",
 
+    "protocolId": "...",
+
     "chainType": "Live" | "Local" | "Development" | { "Custom": "..." },
 
     "bootnodes": [
@@ -42,6 +44,8 @@ The JSON object returned by this function has the following format:
 - `name` is a string containing the name of the chain, identical to the result of `chainSpec_v1_chainName`.
 
 - `id` is a string containing the identifier of the chain.
+
+- `protocolId` is an _optional_ string containing the network protocol id that identifies the chain.
 
 - `chainType` is an object containing the type of the chain.  
 This information can be used by tools to display additional information to the user.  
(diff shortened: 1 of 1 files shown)


lexnv:
In this revision I've added the consensus and finality bits we need to sync with the chain.
Its a bit less elegant than the agnostic checkpoint, but this way smoldot can provide the chainSpec without making queries to the full nodes. 

I suggest adding this as unstable and implementing it in substrate to have a starting point.
We can always go back to the agnostic checkpoint if you think it is not too error prone

Would love your feedback on this
@tomaka @jsdw  🙏

PR #129, opened 2024-02-02
fix: `rpc_methods` return value

josepot:


src/api/rpc_methods.md
```diff
@@ -2,4 +2,4 @@
 
 **Parameters**: *none*
 
-**Return value**: Array of strings indicating the names of all the JSON-RPC functions supported by the JSON-RPC server, including this one.
+**Return value**: A plain JavaScript object with a single entry named "`methods`". This entry is an array of strings indicating the names of all the JSON-RPC functions supported by the JSON-RPC server, including this one.
```
tomaka, line 5: What's a "JavaScript object"? This is spec is not about JavaScript

tomaka: The spec has multiple examples of functions that return an object, see for example https://paritytech.github.io/json-rpc-interface-spec/api/archive_unstable_storage.html

josepot: > What's a "JavaScript object"? This is spec is not about JavaScript

This is a **JSON** RPC spec.

JSON => JavaScript Object Notation

> JSON is built on two structures:

> A collection of name/value pairs. In various languages, this is realized as an object, record, struct, dictionary, hash table, keyed list, or associative array.

> An ordered list of values. In most languages, this is realized as an array, vector, list, or sequence.

josepot: I changed it to match the other "return value" styles. However, I really don't see why you would be against a description that mentions a "JavaScript object", while at the same time being perfectly ok with a description that reads: 

> Return value: a Javascript Object Notation object...

The "beauty" of JSON is that it allows anyone to translate those notated JS objects into whichever data-structure works bes in their language. However, that doesn't change the fact that what it's representing is a JS object.

src/api/rpc_methods.md
```diff
@@ -2,4 +2,16 @@
...
+
+```
+{
+    "methods": [...]
+}
+```
+
+Where:
+
+- `methods` contains an array of strings indicating the names of all the JSON-RPC functions supported by the JSON-RPC server, including this one.
```
tomaka, line 17:
```suggestion
- `methods` contains an array of strings indicating the names of all the JSON-RPC functions supported by the JSON-RPC server, including `rpc_methods` itself.
```

"This one" is a bit less clear now

tomaka (approved):

niklasad1 (approved):
nice

lexnv (approved):

josepot:
Friendly reminder that I can no longer merge this 🙂

Issue #125, opened 2024-01-19
`chainHead_call` (et al): clarify that response from this comes back before any events on the `follow` subscription. 


jsdw:
Raising on behalf of https://github.com/paritytech/polkadot-sdk/issues/2992#issuecomment-1900285167

In our current substrate implementation of these methods, the response from a call to eg `chainHead_call` may arrive back after one or more events associated with it arrive on the corresponding `chainHead_follow` subscription.

The spec doesn't make it clear that we expect a specific order of events (eg, subscription ID comes back on `call` before any events come back on `follow`). Eg on `chainHead_call` it just says of the "started" response:

> This return value indicates that the request has successfully started.
>
> operationId is a string containing an opaque value representing the operation.

@tomaka, do you see any issue with us specifying that we'll always see a response to `call` and related before seeing any corresponding events on the `follow` subscription?

josepot:
> The spec doesn't make it clear that we expect a specific order of events (eg, subscription ID comes back on call before any events come back on follow). Eg on chainHead_call it just says of the "started" response

Perhaps we could make it more explicit. However, I think that anyone reading the spec would immediately understand that the order in which the message arrive is important, and that a notification that represents an event shouldn't be sent before the client has received its identifier. If we need to make this even more explicit, that is fine by me, but please notice that even @lexnv proposed changing the spec to allow the messages to come out of order. So, he clearly also assumed that the spec was setting this expectation.

IMO the "[Bounded Queues](https://paritytech.github.io/json-rpc-interface-spec/dos-attacks-resilience.html#bounded-queues)" section of the spec in combination with the docs of the chainHead group of functions already make this pretty clear.

jsdw:
When I implemented this stuff in Subxt, I'm pretty sure I didn't assume anything about the ordering and did it to handle race conditions (though in Rust I think that is a possibility regardless, so it was always a safer bet). IMO none of those things make the relationship between outputs _different calls_ clear, but I'm not going to argue this point if others think that it was obvious enough.

josepot:
> I'm pretty sure I didn't assume anything about the ordering and did it to handle race conditions 

[...]

PR #137, opened 2024-02-14
chainHead: Allow single connection calls and enforce order of events

lexnv:
To avoid any confusion wrt the order of received events:
- State that the `operationID` is received before any other announcement that may contain the `operationID`
- The `chainHead` subscription ID is only valid in the context of the same established connection
  - This disallows the users from having one `chainHead_follow` connection; then call `chainHead_storage` from another connection (different socket) and drive the events on the first connection. For this to happen, the user would need to know the IP address of the server.

Closes https://github.com/paritytech/json-rpc-interface-spec/issues/125.

### DIFF
--- a/src/api.md
+++ b/src/api.md
@@ -6,6 +6,8 @@ Any missing parameter, or parameter with an invalid format, should result in a J
 
 Any function returning an opaque subscription or operation ID ensures that this is returned before any related notifications are generated.
 
+The functions within each respective category must be called from the same connection in order to work together.
+
 ## Glossary
 
 - "hexadecimal-encoded" designates a binary value encoded as hexadecimal. The value must either be empty, or start with `"0x"` and contain an even number of characters.
--- a/src/api.md
+++ b/src/api.md
@@ -4,7 +4,7 @@ Note that all parameters are mandatory unless specified otherwise. All functions
 
 Any missing parameter, or parameter with an invalid format, should result in a JSON-RPC error being returned, as described in the JSON-RPC specification.
 
-Any function returning an opaque subscription or operation ID ensures that the value is returned before any subsequent notifications are generated.
+Any function returning an opaque subscription or operation ID ensures that this is returned before any related notifications are generated.
 
 ## Glossary
 
--- a/src/api.md
+++ b/src/api.md
@@ -4,6 +4,8 @@ Note that all parameters are mandatory unless specified otherwise. All functions
 
 Any missing parameter, or parameter with an invalid format, should result in a JSON-RPC error being returned, as described in the JSON-RPC specification.
 
+Any function returning an opaque subscription or operation ID ensures that the value is returned before any subsequent notifications are generated.
+
 ## Glossary
 
 - "hexadecimal-encoded" designates a binary value encoded as hexadecimal. The value must either be empty, or start with `"0x"` and contain an even number of characters.
--- a/src/api/chainHead.md
+++ b/src/api/chainHead.md
@@ -6,7 +6,7 @@ The most important function in this category is `chainHead_unstable_follow`. It 
 
 These functions are the functions most of the JSON-RPC clients will most commonly use. A JSON-RPC server implementation is encouraged to prioritize serving these functions over other functions, and to put pinned blocks in a quickly-accessible cache.
 
-In case of a JSON-RPC server that accepts multiple different client connections, the `chainHead` functions must be called from the same connection.
+The `chainHead` functions must be called from the same connection in order to work together.
 
 ## Usage
 
--- a/src/api/chainHead.md
+++ b/src/api/chainHead.md
@@ -6,8 +6,6 @@ The most important function in this category is `chainHead_unstable_follow`. It 
 
 These functions are the functions most of the JSON-RPC clients will most commonly use. A JSON-RPC server implementation is encouraged to prioritize serving these functions over other functions, and to put pinned blocks in a quickly-accessible cache.
 
-The `chainHead` functions must be called from the same connection in order to work together.
-
 ## Usage
 
 _This section contains a small beginner guide destined for JSON-RPC client users._
--- a/src/api/chainHead.md
+++ b/src/api/chainHead.md
@@ -6,6 +6,8 @@ The most important function in this category is `chainHead_unstable_follow`. It 
 
 These functions are the functions most of the JSON-RPC clients will most commonly use. A JSON-RPC server implementation is encouraged to prioritize serving these functions over other functions, and to put pinned blocks in a quickly-accessible cache.
 
+In case of a JSON-RPC server that accepts multiple different client connections, the `chainHead` functions must be called from the same connection.
+
 ## Usage
 
 _This section contains a small beginner guide destined for JSON-RPC client users._
--- a/src/api/chainHead_unstable_body.md
+++ b/src/api/chainHead_unstable_body.md
@@ -43,6 +43,8 @@ The JSON-RPC server must start obtaining the body (in other words the list of tr
 
 The progress of the operation is indicated through `operationBodyDone`, `operationInaccessible`, or `operationError` notifications generated on the corresponding `chainHead_unstable_follow` subscription.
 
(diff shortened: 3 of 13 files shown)


src/api/chainHead_unstable_call.md
```diff
@@ -47,6 +47,8 @@ The JSON-RPC server must invoke the entry point of the runtime of the given bloc
 
 The progress of the operation is indicated through `operationCallDone`, `operationInaccessible`, or `operationError` notifications generated on the corresponding `chainHead_unstable_follow` subscription.
 
+This function can only be called from the same connection as the one that made the `chainHead_unstable_follow` subscription. The `operationId` is guaranteed to be generated before the `operationCallDone`, `operationInaccessible`, or `operationError` notifications are generated on the corresponding `chainHead_unstable_follow` subscription.
```
jsdw, line 50:
```suggestion
This function must be called from the same connection as the corresponding `chainHead_unstable_follow` subscription. The `operationId` is guaranteed to be returned before any operation notifications on the corresponding `chainHead_unstable_follow` subscription.
```

Just trying to condense it down a bit :)

src/api/chainHead_unstable_stopOperation.md
```diff
@@ -12,3 +12,5 @@ Stops an operation started with `chainHead_unstable_body`, `chainHead_unstable_c
 Has no effect if the `followSubscription` is invalid or stale.
 
 JSON-RPC client implementations must be aware that, due to the asynchronous nature of JSON-RPC client <-> server communication, they might still receive notifications about the operation, for example because a notification was already in the process of being sent back by the JSON-RPC server.
+
+This function can only be called from the same connection as the one that made the `chainHead_unstable_follow` subscription.
```
jsdw, line 16:
```suggestion
This function must be called from the same connection as the one that made the `chainHead_unstable_follow` subscription.
```

src/api/chainHead_unstable_continue.md
```diff
@@ -11,6 +11,8 @@ Resumes a storage fetch started with `chainHead_unstable_storage` after it has g
 
 Has no effect if the `operationId` is invalid or refers to an operation that has emitted a `{"event": "operationInaccessible"}` event, or if the `followSubscription` is invalid or stale.
 
+This function can only be called from the same connection as the one that made the `chainHead_unstable_follow` subscription.
```
jsdw, line 14:
```suggestion
This function must be called from the same connection as the one that made the `chainHead_unstable_follow` subscription.
```

jsdw (approved):

src/api/chainHead_unstable_body.md
```diff
@@ -43,6 +43,8 @@ The JSON-RPC server must start obtaining the body (in other words the list of tr
 
 The progress of the operation is indicated through `operationBodyDone`, `operationInaccessible`, or `operationError` notifications generated on the corresponding `chainHead_unstable_follow` subscription.
 
+This function must be called from the same connection as the corresponding `chainHead_unstable_follow` subscription. The `operationId` is guaranteed to be returned before any operation notifications on the corresponding `chainHead_unstable_follow` subscription.
```
tomaka, line 46:
```suggestion
This function must be called from the same connection as the corresponding `chainHead_unstable_follow` subscription. The `operationId` is guaranteed to be returned before any operation notification on the corresponding `chainHead_unstable_follow` subscription.
```

src/api/chainHead_unstable_call.md
```diff
@@ -47,6 +47,8 @@ The JSON-RPC server must invoke the entry point of the runtime of the given bloc
 
 The progress of the operation is indicated through `operationCallDone`, `operationInaccessible`, or `operationError` notifications generated on the corresponding `chainHead_unstable_follow` subscription.
 
+This function must be called from the same connection as the corresponding `chainHead_unstable_follow` subscription. The `operationId` is guaranteed to be returned before any operation notifications on the corresponding `chainHead_unstable_follow` subscription.
```
tomaka, line 50:
```suggestion
This function must be called from the same connection as the corresponding `chainHead_unstable_follow` subscription. The `operationId` is guaranteed to be returned before any operation notification on the corresponding `chainHead_unstable_follow` subscription.
```

src/api/chainHead_unstable_storage.md
```diff
@@ -56,6 +56,8 @@ For the purpose of storage requests, the trie root hash of the child tries of th
 
 The progress of the operation is indicated through `operationStorageItems`, `operationWaitingForContinue`, `operationStorageDone`, `operationInaccessible`, or `operationError` notifications generated on the corresponding `chainHead_unstable_follow` subscription.
 
+This function must be called from the same connection as the corresponding `chainHead_unstable_follow` subscription. The `operationId` is guaranteed to be returned before any operation notifications on the corresponding `chainHead_unstable_follow` subscription.
```
tomaka, line 59:
```suggestion
This function must be called from the same connection as the corresponding `chainHead_unstable_follow` subscription. The `operationId` is guaranteed to be returned before any operation notification on the corresponding `chainHead_unstable_follow` subscription.
```

tomaka:
I'm not such a fan of the word "connection". The JSON-RPC API doesn't necessarily go over a connection.

The fact that the JSON-RPC API is done over WebSocket is more or less out of scope (although, of course, pragmatically speaking we shouldn't design the API in such a way that it wouldn't work on WebSocket).
In principle, given that you're only ever supposed to connect to JSON-RPC API to a node that you own, doing JSON-RPC over inter-process communication or function calls is arguably more idiomatic than over WebSocket.

Additionally, there's no reason why this is explained for `chainHead_follow` but not for `sudo_network_watch` or `transactionsWatch`, even though they're also concerned.

In my opinion, we should only mention this as a note in `chainHead_follow` and other subscriptions, and add "in case of a JSON-RPC server that accepts multiple different clients" as a preamble.

jsdw:
> Wouldn't it be easier to add a general remark in the [API specification](https://paritytech.github.io/json-rpc-interface-spec/api.html#api-specification) section stating that any time that a method returns an opaque value representing a subscription, that opaque value is guaranteed to be returned before any of its upcoming notifications?

This would work for me too!

src/api.md
```diff
@@ -4,6 +4,8 @@ Note that all parameters are mandatory unless specified otherwise. All functions
 
 Any missing parameter, or parameter with an invalid format, should result in a JSON-RPC error being returned, as described in the JSON-RPC specification.
 
+Any function returning an opaque subscription or operation ID ensures that the value is returned before any subsequent notifications are generated.
```
jsdw, line 7: Nit: just felt a touch clearer

```suggestion
Any function returning an opaque subscription or operation ID ensures that this is returned before any related notifications are generated.
```

src/api/chainHead.md
```diff
@@ -6,6 +6,8 @@ The most important function in this category is `chainHead_unstable_follow`. It
 
 These functions are the functions most of the JSON-RPC clients will most commonly use. A JSON-RPC server implementation is encouraged to prioritize serving these functions over other functions, and to put pinned blocks in a quickly-accessible cache.
 
+In case of a JSON-RPC server that accepts multiple different client connections, the `chainHead` functions must be called from the same connection.
```
jsdw, line 9:
```suggestion
The `chainHead` functions must be called from the same connection in order to work together.
```

Do we need the beginning bit, because whether multiple or one conenction allowed, the same is true anyway?

And then I just added "in order to work together." because you can of course call them from different connections (where possible), but in that case they won't work together. There's prob a better way to work this overall (ie add an example of what is meant), but I didn't want to dedicate lots of words to this minor point :)

lexnv: Yep I like this better :D

jsdw (approved):
LGTM! We could prob move the same conneciton text to the general section too but I don't mind!

jsdw (approved):

src/api/chainHead.md
tomaka: I'm the one who asked for this beginning bit in the comment above. The problem is that "connection" really means nothing. You addressed my concern, but then it got changed again and back to saying "connection".

In your head you probably understand that it means "WebSocket connection or similar", but the phrase as it is written now is senseless.

PR #146, opened 2024-03-21
transactionWatch: Limit the number of active subscriptions and clarify error codes

lexnv:
This PR ensures that the server accepts at least 4 `transactionWatch` subscriptions, similar to the `transactionBroadcast` API.

While at it, have defined the error codes returned by this API
- `-38000` - more than 4 `transactionWatch` subscriptions are created (identical error code with chainHead_follow)
- `-32602` - Invalid Parameter from the rpc spec

Have also added a mention wrt invalid decoded transaction bytes in the `Invalid` event.

### DIFF
--- a/src/api/transactionWatch_unstable_submitAndWatch.md
+++ b/src/api/transactionWatch_unstable_submitAndWatch.md
@@ -169,4 +169,4 @@ JSON-RPC servers are allowed to skip sending events as long as it properly keeps
 ## Possible errors
 
 - A JSON-RPC error with error code `-32800` can be generated if the JSON-RPC client has already opened 4 or more `transactionWatch_unstable_submitAndWatch` subscriptions.
-- A JSON-RPC error is generated if the `transaction` parameter has an invalid format, but no error is produced if the bytes of the `transaction`, once decoded, are invalid. Instead, an `invalid` notification will be generated.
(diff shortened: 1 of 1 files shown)


src/api/transactionWatch_unstable_submitAndWatch.md
```diff
@@ -165,4 +168,5 @@ JSON-RPC servers are allowed to skip sending events as long as it properly keeps
 
 ## Possible errors
 
-A JSON-RPC error is generated if the `transaction` parameter has an invalid format, but no error is produced if the bytes of the `transaction`, once decoded, are invalid. Instead, an `invalid` notification will be generated.
+- A JSON-RPC error with error code `-32800` can be generated if the JSON-RPC client has already opened 4 or more `transactionWatch_unstable_submitAndWatch` subscriptions.
+- A JSON-RPC error with error code `-32602` is generated if the `transaction` parameter has an invalid format, but no error is produced if the bytes of the `transaction`, once decoded, are invalid. Instead, an `invalid` notification will be generated.
```
jsdw, line 172:
```suggestion
- A JSON-RPC error with error code `-32602` is generated if the `transaction` parameter has an invalid format. Note that no error is produced if the bytes of the `transaction`, once decoded, are invalid. Instead, an `invalid` notification will be generated.
```

Just a small nit on the wording

jsdw (dismissed):
LGTM; is consistent with chainHead_follow :)

jsdw (approved):

tomaka:
This can't be guaranteed, because transactions stay in the pool when a JSON-RPC client disconnects.

PR #147, opened 2024-03-21
Stabilize `chainHead` to version 1

lexnv:
This PR stabilizes all chainHead methods to version 1.


jsdw (approved):
Looks good to me! We've been using these methods for a while in Subxt and they are roughly at parity with the legacy methods in terms of the functionality exposed.

My stance is that, although there are likely a few changes I'd like to see here, it's much better to stabilise _something_ now rather than get caught up with aiming for perfection and not get anything stabilised for a while. 

Once these are stabilised, we can introduce v2 etc methods if needed to accomodate future requests and changes. We can also have confidence that PAPI will Just Work, and work to stabilise the "unstable backend" in Subxt (which relies on these currently unstable methods).

tomaka (approved):
👍

PR #148, opened 2024-04-03
Revert "Stabilize `transactionWatch` to version 1"

tomaka:
Reverts paritytech/json-rpc-interface-spec#144

https://github.com/paritytech/json-rpc-interface-spec/pull/146 is flawed. This shouldn't have been stabilized.


lexnv (approved):
LGTM!

Just to double-check, we are reverting this because the imposed limit of having at least 4 subscriptions accepted per connection cannot be guaranteed once the transaction enters the pool.

Therefore, we might have situations where the `transctionWatch` is capped to 4 subscriptions although the same user has more than 4 transactions currently active in the pool.

jsdw (dismissed):

jsdw:
I resolved the weird conflict in #149 since I can't easily edit fork PRs :)

PR #152, opened 2024-04-15
transactionWatch: Remove limits and stabilize to version 1

lexnv:
This PR is a followup of the proposal from: https://github.com/paritytech/json-rpc-interface-spec/pull/150#pullrequestreview-1988787293

In this PR, the concept of limits is removed from the `transactionWatch`, as noted in: https://github.com/paritytech/json-rpc-interface-spec/pull/150.
- Servers can still return `Transaction::Dropped` if the node cannot keep up with the number of transactions (as per spec), however servers are not required to support a minimum of 4 subscriptions per client
- TransactionWatch is stabilized to version 1

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -44,5 +44,5 @@
     - [transaction_v1_broadcast](api/transaction_v1_broadcast.md)
     - [transaction_v1_stop](api/transaction_v1_stop.md)
   - [transactionWatch](api/transactionWatch.md)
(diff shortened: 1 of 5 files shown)


tomaka:
I don't really understand the need to do these two widely different things in a single PR?

lexnv:
That makes sense, thanks for the review! 🙏 

Have broken this down into:
- https://github.com/paritytech/json-rpc-interface-spec/pull/153 
- https://github.com/paritytech/json-rpc-interface-spec/pull/154

PR #153, opened 2024-04-17
txWatch: Remove the txWatch limits

lexnv:
This PR is a followup of the proposal from: https://github.com/paritytech/json-rpc-interface-spec/pull/150#pullrequestreview-1988787293

In this PR, the concept of limits is removed from the `transactionWatch`, as noted in: https://github.com/paritytech/json-rpc-interface-spec/pull/150.
- Servers can still return `Transaction::Dropped` if the node cannot keep up with the number of transactions (as per spec), however servers are not required to support a minimum of 4 subscriptions per client


jsdw (approved):

tomaka:
So this PR is identical to #150

tomaka (approved):

niklasad1 (approved):

Issue #108, opened 2023-10-16
RPC to return storage diff between start_block and end_block


dvcpull:
This rpc will allow to quickly fetch the list of keys that have been changed/deleted between two different blocks, this is useful for chain analysis tools like an indexer and can then be used for processing.

The api would look like :

```rust
/// Returns a storage diff between given block and previous block
#[method(name = "state_getStorageDiff", aliases = ["state_getStorageDiffAt"], blocking)]
fn storage_diff(&self, block: Hash, included_prefixes: Option<Vec<StorageKey>>, excluded_prefixes: Option<Vec<StorageKey>>, include_modified_child_tries: Option<bool>) -> RpcResult<(StorageCollection, Option<ChildStorageCollection>)>;

```

Example : 
`curl -sS -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "state_getStorageDiff", "params": ["0xaa254e5ff33843f4abfb2527c7273239f395f7f7da10b6018646f4ae7257f913", "0xb2990969fd7771959f534d3d985603b5632fb6a3faf628d7ba79c564424de7ae"]}' http://localhost:9944/`

Result : 
```
[...]

PR #159, opened 2024-08-01
archive: Introduce archive_storageDiff for indexers

lexnv:
This PR introduces a new unstable method to the archive class.

The method `archive_unstable_storageDiff` aims to help indexers in providing storage differences between two blocks.

By default, the method returns the difference between `currentHash` and its parent.

Furthermore, users have the ability to specify the keys of interest:
- prefixes: A list of key prefixes under which the users will receive notifications
- exclude key prefixes: A list of key prefixes to exclude from the results
- childTrie: Useful for computing the difference for a child trie instead of the main storage trie

For each individual prefix, the users can specify if they need:
- `value`: Returns the value of the key from the `currentHash` block
- `hash`: Similar to value but returns the hash
- `none`: Users are only interested to see if the key was modified, added or removed.

For symmetry with the chainHead functions:
- a `storageDiffStop` method is used to stop the subscription
- a `storageDiffContinue` method is used for resuming storage differences

Closes: https://github.com/paritytech/json-rpc-interface-spec/issues/108

cc @paritytech/subxt-team @tomaka @jsdw @josepot


src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,114 @@
+# archive_unstable_storageDiff
+
+**Parameters**:
+
+- `currentHash`: String containing a hexadecimal-encoded hash of the header of the block whose storage difference will be retrieved. The storage difference is calculated between the `currentHash` block and the parent of the `currentHash` block.
+
+- `fromHash` (optional): String containing a hexadecimal-encoded hash of the header of the block from which the storage difference will be calculated. When this parameter is provided, the storage difference is calculated between the `currentHash` block and the `fromHash` block. If this parameter is not provided, the storage difference is calculated between the `currentHash` block and the parent of the `currentHash` block.
```
jsdw, line 7: Nit: I wonder about the names. Perhaps `hash` for the hash to calculate the diff of, and `previousHash` for the optional previous hash (using prev block if not given)?

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,114 @@
...
+        ],
+        "childTrie": "0x...",
+    }
+    ```
+
+  If the `items` parameter is not provided, the storage difference is calculated for all keys of the main storage trie.
+  - `prefixes` (optional): Array of JSON objects describing how the storage difference will be calculated. Each object contains the following fields:
+    - `key`: String containing the hexadecimal-encoded key prefix under which the storage difference is calculated.
+    - `type`: String equal to one of: `value`, `hash`, `none`.
+  - containing the key prefixes for which the storage difference will be calculated. If this parameter is not provided, the storage difference is calculated for all keys.
```
jsdw, line 30: Is this line leftover by mistake? It doesn't appear to fit in :)

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,114 @@
...
+  - `excludeKeyPrefixes` (optional): Array of strings containing the key prefixes for which the storage difference will not be calculated. If this parameter is not provided, the storage difference is calculated for all keys.
```
jsdw, line 31: Is there a use case in mind for `excludeKeyPrefixes`? If not then I wonder whether it hsould be left out of this version :)

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,114 @@
...
+
+**Return value**: String containing an opaque value representing the operation.
+
+This function calculates the storage difference between two blocks. The storage difference is calculated by comparing the storage of the `fromHash` block with the storage of the `currentHash` block. If the `fromHash` parameter is not provided, the storage difference is calculated between the parent of the `currentHash` block and the `currentHash` block.
+
+The storage difference is calculated for the main storage trie by default. The `items` parameter can be used to specify the key prefixes for which the storage difference will be calculated. The `prefixes` field contains an array of objects, each describing a key prefix and the type of the storage difference to calculate. The `excludeKeyPrefixes` field contains an array of key prefixes for which the storage difference will not be calculated. The `childTrie` field specifies the child trie for which the storage difference will be calculated.
```
jsdw, line 38: THese bits feel a bit duplicated, ie after reading the parameters stuff I already know these things. Perhaps it's worth a bit of duplication but unsure how much to keep!

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,114 @@
...
+            "hash": "0x...",
+            "type": "added" | "modified" | "deleted",
+        },
+    ],
+}
+```
+
+The `differences` field contains an array of objects, each describing a storage difference.
+
+The `key` field contains the hexadecimal-encoded key of the storage entry. If the a key prefix was provided in the `items` parameter, the `key` field is guaranteed to start with one of the key prefixes provided. If the `excludeKeyPrefixes` field was provided in the `items` parameter, the `key` field is guaranteed not to start with any of the excluded key prefixes provided.
```
jsdw, line 77:
```suggestion
The `key` field contains the hexadecimal-encoded key of the storage entry. If the key prefix was provided in the `items` parameter, the `key` field is guaranteed to start with one of the key prefixes provided. If the `excludeKeyPrefixes` field was provided in the `items` parameter, the `key` field is guaranteed not to start with any of the excluded key prefixes provided.
```

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,114 @@
...
+
+### continue
+
+```json
+{
+    "event": "continue",
+}
+```
+
+This notification is sent when the storage difference calculation is paused and needs to be resumed. The user must call `archive_unstable_storageDiffContinue` with the subscription value to continue the calculation.
```
jsdw, line 99: `archive_unstable_continue` would more closely align with `chainHead_v1_continue`, unless there's a reason to be more specific here?

src/api/archive_unstable_storageDiffStop.md
```diff
@@ -0,0 +1,11 @@
+# archive_unstable_storageDiff
```
jsdw, line 1: misnamed title :)

But also; worth calling it just `archive_unstable_stop`?

src/api/archive_unstable_storageDiffContinue.md
```diff
@@ -0,0 +1,13 @@
+# archive_unstable_storageDiffContinue
```
jsdw, line 1: I wonder about calling it `archive_unstable_continue` to align with the chainHead one?

jsdw:
I wonder about the performance implications of eg not providing prefixes (or providing many or large ones), but think that the streaming makes sense as a way of at least mitigating the performance hit of this call on the node! As an API though, off the top of my head it looks solid and feels like it would cover the use cases I can think of!

src/api/archive_unstable_storageDiff.md
tomaka, line 7: What happens if `fromHash` isn't an ancestor of `currentHash`?

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,114 @@
+# archive_unstable_storageDiff
+
+**Parameters**:
+
+- `currentHash`: String containing a hexadecimal-encoded hash of the header of the block whose storage difference will be retrieved. The storage difference is calculated between the `currentHash` block and the parent of the `currentHash` block.
```
tomaka, line 5: The second sentence is contradicting the `fromHash` parameter.

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,114 @@
...
+            "0x...",
+        ],
+        "childTrie": "0x...",
+    }
+    ```
+
+  If the `items` parameter is not provided, the storage difference is calculated for all keys of the main storage trie.
+  - `prefixes` (optional): Array of JSON objects describing how the storage difference will be calculated. Each object contains the following fields:
+    - `key`: String containing the hexadecimal-encoded key prefix under which the storage difference is calculated.
+    - `type`: String equal to one of: `value`, `hash`, `none`.
```
tomaka, line 29: "Calculating under a prefix" doesn't mean anything to me.
The "type" of a "prefix" doesn't make sense to me either.

src/api/archive_unstable_storageDiff.md
tomaka, line 31: You don't "calculate" a storage diff. You probably mean that you exclude some keys from the result sent to the JSON-RPC client.

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,114 @@
...
+  - containing the key prefixes for which the storage difference will be calculated. If this parameter is not provided, the storage difference is calculated for all keys.
+  - `excludeKeyPrefixes` (optional): Array of strings containing the key prefixes for which the storage difference will not be calculated. If this parameter is not provided, the storage difference is calculated for all keys.
+  - `childTrie` (optional): A string containing the hexadecimal-encoded key of the child trie of the "default" namespace. If this parameter is not provided, the storage difference is calculated for the main storage.
```
tomaka, line 32: I guess that you just put this parameter here without giving much thoughts to it in order to get child tries out of the way, but it is very incoherent with the rest of the design of the function.
The function lets you query the difference between a range of blocks, but then you would have to query each child trie one by one? That's extremely weird to me.

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,114 @@
...
+
+**Return value**: String containing an opaque value representing the operation.
```
tomaka, line 34: For the sake of simplicity, I've been trying to avoid subscriptions for all `archive`-related functions.
The subscription functions are important for `chainHead`, because you want be able to cancel operations in order to keep bandwidth usage to a minimum, and because you want to be able to cancel operations related to blocks that end up not being canonical.
However, none of these two things apply here (and don't apply for any `archive`-prefix function), so I don't really see the point of having a subscription.

src/api/archive_unstable_storageDiff.md
tomaka: Also I don't think that the naming is appropriate. What exactly is "current" here?

tomaka, line 5: Also I don't think that the naming is appropriate. What exactly is "current" here, given that the block can be very old?

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,114 @@
...
+
+```json
+{
+    "event": "done",
+}
+```
+
+This notification is sent when the storage difference calculation is complete.
+
+## Possible errors
```
tomaka, line 111: What if `currentHash` or `fromHash` are unknown block hashes?

src/api/archive_unstable_storageDiff.md
jsdw: I assumed that the point of the subscription here is that calculating the diffs can be quite slow, so by streaming them (there can be a bunch of prefixes asked for) the user can `stop` them (and be asked to `continue`) if they find what they need or decide they don't need the server to continue working?

I agree though that it'd be much nicer to avoid the subscriptions though, so if this isn't an issue then it'd be great to simplify it!

tomaka: > if they find what they need or decide they don't need the server to continue working?

In my opinion it is a better idea for the client to send queries progressively, for example query from block 1 to 100, then block 101 to 200, then block 201 to 300, etc.
It is not so different to do so compared to sending a "continue", except that the client has more control.

src/api/archive_unstable_storageDiff.md
lexnv: Yep I like that naming better, thanks!

src/api/archive_unstable_storageDiff.md
lexnv: Have removed the `excludeKey` since it added some extra complexity, will add it back if users feel the need for this 👍 thanks

src/api/archive_unstable_storageDiff.md
lexnv: Made the subscription a method after all, thanks

src/api/archive_unstable_storageDiffStop.md
lexnv: Made the subscription a method after all, thanks

src/api/archive_unstable_storageDiffContinue.md
lexnv: Made the subscription a method after all, thanks

src/api/archive_unstable_storageDiff.md
lexnv: Nice catch! Have added a section to specify that the `previousHash` (old `fromHash`) must be an ancestor of the provided hash, thanks!

src/api/archive_unstable_storageDiff.md
lexnv: Have removed the second sentance and provided the details in `previousHash` section, thanks!

src/api/archive_unstable_storageDiff.md
lexnv: Have remove the `excludeKeyPrefixes` to simplify the API a bit and have adjusted the `key` field to specify that `Only the storage entries whose key starts with the provided prefix are returned.`, thanks!

src/api/archive_unstable_storageDiff.md
lexnv: I do like this approach! Turning this subscription into a method is easier to follow from the user's perspective. I'll do some performance measurements for storage differences once we implement this in Substrate.

We could probably add a disk cache to make queries a bit faster if that turns to be a performance issue

src/api/archive_unstable_storageDiff.md
lexnv: Yep that makes sense, have applied James suggestion of naming `currentHash` -> `hash` and `fromHash` -> `previousHash` 🙏

src/api/archive_unstable_storageDiff.md
lexnv: Have added a new sentace to describe this a bit better: " Only the storage entries whose key starts with the provided prefix are returned." , thanks!

src/api/archive_unstable_storageDiff.md
lexnv: Yeap, I wondering as well if that's the right place to put it. Have extended it similar to:

```
items [
 {
    (optional) prefixes...
    trieType: mainTrie | childTrie
    (optional) childTrieKey
 } 
]
```

This should provide the storage difference of multiple tries. 
I've added the nonoptional `trieType` because it felt odd for users to say:
` items: [ { }, { "childKey": ...}]`  (ie have an empty object for main trie queries)

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,98 @@
...
+            "prefixes": [
+                {
+                    "key": "0x...",
+                    "type": "value" | "hash" | "none",
+                },
+            ],
+
+            "trieType": "mainTrie" | "childTrie",
+            "childTrieKey": "0x...",
+        },
```
jsdw, line 23: It feels a little off to me that we have two arrays nested inside eachother in order to presumably support child tree access. I wonder whether we could simplify itms to be:

```json
"items": [
    {
        "key": "0x...",
        "type": "value" | "hash" | "none",
        "childTrieKey": "0x...",
    }
]
```

If "childTrieKey" is provided then we try accessing said child trie, else we assume we're trying to access the main trie (which I guess is the overwhelming default for most, but not 100% sure!).

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,98 @@
...
+    - `childTrie`: The storage difference is calculated for the child trie of the "default" namespace. If this type is provided, the `childTrieKey` field must be provided.
+
+  - `childTrieKey`: String containing the hexadecimal-encoded key of the child trie of the "default" namespace. This field is only present when the `trieType` field is set to `childTrie`.
+
+**Return value**: A JSON object.
+
+The JSON object returned by this function has the following format:
+
+```json
+{
```
jsdw, line 46: If we did the above, we'd also tweak this to be a single array whose entries contained a "childTrieKey" if one was provided in the input for that key ie

```json
"result": [
        {
            "key": "0x...",
            "value": "0x...",
            "hash": "0x...",
            "type": "added" | "modified" | "deleted",
            "childTrieKey": "0x..." | null,
        },
]
```

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,98 @@
...
+        }
+    ],
+}
+```
+
+The `result` field contains an array of objects, each containing a JSON object:
+
+The `differences` field is an array of objects describing storage differences. Each element contains the following fields:
+
+- `key`: String containing the hexadecimal-encoded key of the storage entry. If the key prefix was provided in the `items` parameter, the `key` field is guaranteed to start with one of the key prefixes provided.
```
jsdw, line 69: > If the key prefix was provided in the `items` parameter, the `key` field is guaranteed to start with one of the key prefixes provided.

What if the key prefix was not provided in the `items` parameter?

Or perhaps this is aiming for something more like: "String containing the hexadecimal-encoded key of the storage entry. A prefix of this key will have been provided in the `items` input"? I might not 100% understand!

src/api/archive_unstable_storageDiff.md
jsdw: Ah, I see that "prefixes" is optional. If it's not provided, is the point that this method will provide _all_ chanegs in the whole trie?

So either this suggestion wouldn't allow that, or one could set a key like "0x" (or not provide one) to indicate that we are searching for diffs under everything in that trie?

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,98 @@
...
+
+- `items` (optional): Array of objects. The structure of these objects is found below.
+
+    ```json
+    "items": [
+        {
+            "prefixes": [
+                {
+                    "key": "0x...",
+                    "type": "value" | "hash" | "none",
```
niklasad1, line 17: I find `"none"` hard to understand and AFAIU it's a way to get all storage diffs and ignore the type of the storage diff.

I suggest rename it to "all" or something like that.

Can you document the "type" options a bit more through? I didn't see any explanation of "none" except in the PR description.

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,98 @@
...
+The JSON object returned by this function has the following format:
+
+```json
+{
+    "result": [
+        {
+            "differences": [
+                {
+                    "key": "0x...",
+                    "value": "0x...",
```
niklasad1, line 52: Isn't "hash" and "key" mutually exclusive by when you can only enable `type = "hash" | "value" | "none` in the call?

Maybe I'm just don't understand this...

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,98 @@
...
+
+The `trieType` field is a string indicating the trie type for which the storage difference was calculated. The possible values are `mainTrie` and `childTrie`. These have the same meaning as in the `items` parameter.
+
+The `childTrieKey` field is a string containing the hexadecimal-encoded key of the child trie of the "default" namespace. This field is only present when the `trieType` field is set to `childTrie`.
+
+## Overview
+
+This function calculates the storage difference between two blocks. The storage difference is calculated by comparing the storage of the `previousHash` block with the storage of the `hash` block. If the `previousHash` parameter is not provided, the storage difference is calculated between the parent of the `hash` block and the `hash` block.
+
+The JSON-RPC server is encouraged to accept at least one `archive_unstable_storageDiff` subscription per JSON-RPC client. Trying to open more might lead to a JSON-RPC error when calling `archive_unstable_storageDiff`. The JSON-RPC server must return an error code if the server is overloaded and cannot accept new subscriptions.
```
niklasad1, line 88: Subscription? If I read the description properly this just makes a method call with the storageDiff as return value

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,98 @@
...
+
+Users that want to obtain the storage difference between two blocks should use this function instead of calling `archive_unstable_storage` for each block and comparing the results.
+When users are interested in the main trie storage differences, as well as in a child storage difference, they can call this function with `items: [ { "trieType": "mainTrie" }, { "trieType": "childTrie", "childTrieKey": "0x..." } ]`.
+
+## Possible errors
+
+- A JSON-RPC error can be generated if the JSON-RPC client has to many active subscriptions to `archive_unstable_storageDiff`.
```
niklasad1, line 95: same here, subscription?

niklasad1 (commented):
I'm not that familiar with the storage and child storage stuff to understand why the API actually needs to expose whether the storage keys/prefixes is a main or child storage.

It's quite complicated API to use in my opinion but most likely required for reason that I don't understand but I would be much happier to just have not expose "trieType" and "childTrieKey" and let node just inspect the filter of the storage keys.

```json
"items": [
    {
                "key": "0x1.",
                "type": "value" | "hash" | "none",
    }
    {
                "key": "0x2",
                "type": "value" | "hash" | "none",
    }
]
```

In the same manner, make "items" and "trieType" mutually exclusive.

Such one can query: 
1.  [some storage keys],  
2. main storage
3. child storage.

Then also I would be happy to make "type" optional and let the node display the key i.e, "hash" or "value"
I haven't looked at the implementation but just my comments of the overall API.

jsdw:
@lexnv just bumping this incase it fell off the radar! It'd be good to address the comments and get it through review so we can implement it, but I know you're prob busy so not urgent :)

src/api/archive_unstable_storageDiff.md
lexnv: Yep that makes sense! While at it, I've made `key` and `childTrie` optional

src/api/archive_unstable_storageDiff.md
lexnv: Yep the suggestion clarifies things, thanks!

src/api/archive_unstable_storageDiff.md
lexnv: Opps, that was added before, thanks for catching this 🙏

src/api/archive_unstable_storageDiff.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Overview
+
+This function calculates the storage difference between two blocks. The storage difference is calculated by comparing the storage of the `previousHash` block with the storage of the `hash` block. If the `previousHash` parameter is not provided, the storage difference is calculated between the parent of the `hash` block and the `hash` block.
+
+The JSON-RPC server is encouraged to accept at least one `archive_unstable_storageDiff` method per JSON-RPC client. Trying to make more calls might lead to a JSON-RPC error when calling `archive_unstable_storageDiff`. The JSON-RPC server must return an error code if the server is overloaded and cannot accept new method call.
+
+Users that want to obtain the storage difference between two blocks should use this function instead of calling `archive_unstable_storage` for each block and comparing the results.
+When users are interested in the main trie storage differences, as well as in a child storage difference, they can call this function with `items: [ { "trieType": "mainTrie" }, { "trieType": "childTrie", "childTrieKey": "0x..." } ]`.
```
jsdw, line 71: `trieType` was removed, so this sentence needs updating :)

jsdw (approved):
Looks good to me! Just one sentence I spotted htat needed a wee update after other changes :)

jsdw (approved):

niklasad1 (approved):

lexnv:
@tomaka Would love to get your thoughts on this, thanks 🙏

jsdw:
So this looks good, but I'm still hesitant about not using a subscription here. 

If a client asks for a storage diff that is particularly large (ie between two quite distant blocks) I imagine it could be expensive to compute (in terms of the size of response needed as well as time taken to build it). Would this open the node to DoS attacks (clients asking for storage diffs known to be huge)? Or, to avoid this, would some storage diffs simply lead to an error because they are too expensive to compute and so the node refuses?

With a subscription based approach we could do something like progressively streaming the diffs we see as we work from first block to last block. The server wouldn't have to accumulate so much in one go that way, and could also take advantage of backpressure to kill connections to clients that were too slow to receive data back or whatever.

@tomaka I think you preferred a non subscription based approach so perhaps you have an idea already on how to avoid these issues (or maybe don't think that there is any issue in the first place) without needing subscriptions?

lexnv:
Merging this to focus a bit on the implementation, we'll re-evaluate the subscription-based approach after we extract some performance metrics from the new RPC methods, thanks everyone for the feedback and review 🙏 

Please feel free to open any issues / PRs to further discuss this, thanks again!

FleetHub1988 (commented):
Approve 4 merge.

Issue #160, opened 2024-10-10
archive/storageDiff: Suggestion to make storageDiff subscription-based


lexnv:
At the moment, `storageDiff` is a plain method. 

To implement `storageDiff` of two arbitrary blocks, we'll have to iterate through the keys of both blocks and access the storage of those keys requested by the user.

This operation might be expensive enough to DoS the node, considering users might request differences between `genesis` <-> `latest-finalized`.

Another benefit of having a subscription-based approach for `storageDiff` is that we can leverage the backpressure system implemented for the chainHead subscription. 
On the other side, users will have to adapt their code to work with a subscription-based approach. 

Implementation ref: https://github.com/paritytech/polkadot-sdk/pull/5997/files#diff-ac650a2e540d8d37576d05083cce78260192944382ecf1e68f0db50b5d86907eR350-R369

// cc @paritytech/subxt-team @tomaka @josepot Would love to get your thoughts on this 🙏

tomaka:
There are big differences between `chainHead_storage` and `archive_storage`.
In the case of `chainHead`, a light client needs to send multiple networking requests to full nodes, and thus it is advantageous to send notifications to the RPC client progressively. It is also important for the request to be cancelable because it might take a long time, and during this time the RPC client might no longer be interested in the operation. Light clients can't implement `archive_storage` by design, so it's not a problem.

`chainHead_storage` does *not* require/have any backpressure system other than what all JSON-RPC functions have. The `chainHead_storage_continue` function exists not for backpressure, but for the situation where a RPC client is no longer interested in an operation. Without `storage_continue`, in the time between when the client is no longer interested in an operation and when the server receives the `chainHead_stopOperation` the server might have sent several megabytes. `storage_continue` exists so that the server doesn't waste bandwidth.

In theory, `archive_unstable_storage` should be able to simply its result onto the TCP connections, and the back-pressure can be automatically handled. Think for example of a `std::io::Read` or `AsyncRead` trait implementation that generates results on-the-fly when you read it.

> The implementation for the method response needs to have some kind of limit (or at least polkadot-sdk has that) to limit the number of storage queries to avoid DOS attempts.

This is already covered by the function:

```
[...]

PR #161, opened 2024-11-04
archive: Use subscriptions for `storageDiff` 

lexnv:
This PR modifies the `archive_unstable_storageDiff` as a subscription instead of a method.

The change mainly leverages subscriptions' backpressure without implementing a low-level API by the RPC crates. 

This is a followup from: 
- https://github.com/paritytech/json-rpc-interface-spec/issues/160

cc @paritytech/subxt-team

### DIFF
--- a/src/api/archive_unstable_stopStorageDiff.md
+++ b/src/api/archive_unstable_stopStorageDiff.md
@@ -1,0 +1,11 @@
+# archive_unstable_stopStorageDiff
+
+**Parameters**:
+
+- An opaque string that was returned by `archive_unstable_StorageDiff`.
+
+**Return value**: *null*
+
+Stops a subscription started with `archive_unstable_StorageDiff`. Has no effect if the opaque string is invalid or refers to a subscription that has already emitted a `{"event": "storageDiffDone"}` event.
+
+JSON-RPC client implementations must be aware that, due to the asynchronous nature of JSON-RPC client <-> server communication, they might still receive chain updates notifications, for example because these notifications were already in the process of being sent back by the JSON-RPC server.
--- a/src/api/archive_unstable_stopStorageDiff.md
+++ b/src/api/archive_unstable_stopStorageDiff.md
@@ -2,7 +2,7 @@
 
 **Parameters**:
 
-- An opaque string that was returned by `archive_unstable_StorageDiff`.
+- An opaque string that was returned by `archive_unstable_storageDiff`.
 
 **Return value**: *null*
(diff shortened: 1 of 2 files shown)


pkhry (dismissed):

src/api/archive_unstable_storageDiff.md
```diff
@@ -28,25 +28,41 @@
...
+    "params": {
+        "subscription": "...",
+        "result": ...
+    }
+}
+```
+
+Where `subscription` is the value returned by this function, and `result` can be one of:
+
+### storageDiff
```
jsdw, line 50: Perhaps just remove this line, since this object is no longer returned by this function?

jsdw (approved):
LGTM, just a small wording tweak suggestion.

I would still like to see whether @tomaka has any further thoughts, but my general feeling is that this change is worth doing.

I also wonder how consistent this should be with `archive_storage`. That call contains a `paginationStartKey` param which I don't fully understand (given it appears to return everything at once). Should it also be a subscription so that it's able to stream the possibly large set of values found from a users request (again, rather than building up such a set in memory and then sending the whole thing)?

src/api/archive_unstable_stopStorageDiff.md
```diff
@@ -0,0 +1,11 @@
+# archive_unstable_stopStorageDiff
+
+**Parameters**:
+
+- An opaque string that was returned by `archive_unstable_StorageDiff`.
+
+**Return value**: *null*
+
+Stops a subscription started with `archive_unstable_StorageDiff`. Has no effect if the opaque string is invalid or refers to a subscription that has already emitted a `{"event": "storageDiffDone"}` event.
```
niklasad1, line 9:
```suggestion
Stops a subscription started with `archive_unstable_storageDiff`. Has no effect if the opaque string is invalid or refers to a subscription that has already emitted a `{"event": "storageDiffDone"}` event.
```

src/api/archive_unstable_storageDiff.md
lexnv: I've removed the `The JSON object returned by this function has the following format`. Dq: is that the line mentioned? 🤔

niklasad1 (dismissed):

jsdw (dismissed):
This LGTM modulo moving the itesm description back under the `items` bulelt point (https://github.com/paritytech/json-rpc-interface-spec/pull/161#discussion_r1843602371) :)

PR #162, opened 2024-11-15
archive: Use subscriptions for archive_storage

lexnv:
This PR refactors `archive_storge` to be a subscription instead of a method. 

cc @paritytech/subxt-team


jsdw (approved):
This all looks good to me!

PR #183, opened 2026-03-24
Add `bitswap_v1_get` method

dmitry-markin:
This PR standardizes `bitswap_v1_get` RPC used to download data from Polkadot Bulletin Chain. Calling this method is equivalent to downloading the data from a full node via Bitswap protocol. Both light & full nodes must provide this endpoint.

## Follow-ups

`bitswap_v1_stream` method for requesting streaming of multiple Bitswap chunks (multiple CIDs) will be introduced in a separate PR.

CC @ERussel


lexnv (dismissed):

x3c41a (dismissed):

x3c41a (approved):

lexnv (approved):
LGTM! Thanks @dmitry-markin for spec here 🙏

PR #185, opened 2026-04-23
feat(statement): add spec

carlosala:
Add specification for Statement Store API.
This API is meant to be implemented (not only) by:

- Full nodes (e.g. Polkadot-SDK node)
- Light clients (e.g. Smoldot)


carlosala:
This is an alternative version to #184, broader and more robust.

Copying relevant people.
cc @voliva @josepot @alexggh @bkchr @lexnv

src/api/statement.md
```diff
@@ -0,0 +1,19 @@
+# Introduction
+
+The `statement` functions expose the statement store of the node associated to the JSON-RPC server.
+
+Statement stores are weakly coherent. Different JSON-RPC servers can know different subsets of statements at any moment, and no guarantee is offered as to the time it takes for a statement to become visible everywhere.
+
+The `statement_unstable_follow` function creates the main source of notifications. Filters are later attached with successful calls to `statement_unstable_add`, and all notifications for all filters attached to the same follow are generated on that one follow subscription. If the server can't accept a filter on a follow, `statement_unstable_add` returns `{"result": "limitReached"}`. The `statement_unstable_submit` function is independent from follows and can be called at any time.
```
alexggh: What do you say if we  call `statement_unstable_follow`, statement_unstable_subscribe and the return is a subscription, identified by a subscriptionId ?

I don't know, `a follow` reads somehow weirdly to me, maybe it made sense in the context of following a chain, but here the statements are kind of independent of each other.

src/api/statement_unstable_add.md
```diff
@@ -0,0 +1,40 @@
+# statement_unstable_add
```
alexggh:
```suggestion
# statement_unstable_add_filter
```

Maybe ?

src/api/statement_unstable_add.md
```diff
@@ -0,0 +1,40 @@
...
+
+```json
+{
+  "result": "limitReached"
+}
+```
+
+The string returned when this function succeeds is opaque and its meaning can't be interpreted by the JSON-RPC client. It is only unique in the context of the `followSubscription` passed to this function, and is only meant to be passed later to `statement_unstable_remove` together with that same `followSubscription`.
+
+When this function succeeds, the JSON-RPC server attaches the given filter to the follow, returns a `filterId`, and starts replaying on the corresponding `statement_unstable_follow` subscription the statements that are already present in the store and match this filter. These replayed statements are reported through `replayStatements` notifications carrying that `filterId`. If the follow remains alive and the filter isn't removed, the server later emits a `replayDone` notification carrying that same `filterId`, even if no statement matched.
```
alexggh: There is nothing special between statements marked with  `replayStatements` and statements marked with `newStatements`, the clients should not distinguish between them, for example by connecting to two different nodes you could see them differently, so I'm not sure we should give any reply guarantees, other than the fact that at least this amount of statements will be sent.

Additionally, this  also complicates the implementation, since you suggested no newStatements should be sent before replayStatements.

The `remaining` proposed here https://github.com/paritytech/json-rpc-interface-spec/pull/184/changes, it was there just to give a clients a best effort hints of a minimum statements that will be sent, and was preferred by [ERussel](https://github.com/ERussel), as making the implementation on clients easier, rather than having an explicit `replayDone` event.

Why do you think is important to separate them in two different event types  ?

src/api/statement_unstable_remove.md
```diff
@@ -0,0 +1,18 @@
+# statement_unstable_remove
```
alexggh:
```suggestion
# statement_unstable_remove_filter
```

src/api/statement_unstable_unfollow.md
```diff
@@ -0,0 +1,13 @@
+# statement_unstable_unfollow
```
alexggh:
```suggestion
# statement_unstable_unsubscribe
```

src/api/statement_unstable_follow.md
```diff
@@ -0,0 +1,154 @@
...
+
+**Return value**: String containing an opaque value representing the subscription.
+
+This function lets the JSON-RPC client follow statements known to the statement store.
+
+This function works as follows:
+
+- When called, returns an opaque `followSubscription` that is later used to match notifications and to call `statement_unstable_add`, `statement_unstable_remove`, or `statement_unstable_unfollow`.
+
+- The follow starts with no filters attached. As such, this function doesn't immediately generate notifications.
```
alexggh: What would it be wrong to actually allow passing a filter at the creation of the subscription, which would make basically the call of this API equivalent in usage with what we already have implemented ?

alexggh (commented):
@carlosala Thank you for this.

I had a pass overall, I don't think we are that far away, I already mentioned previously that I think this functionality for updating the filters of a subscription is something we would want in the future, although my ideal path would have been with fewer divergences from to the existing implementation.

I left you a few comments, once we reach a consensus on those, I think the path forward could be:

1. We consider this the next version of statement-store RPC.
2. We have polkadot-sdk node and smoldot implement this version alongside the existing implementation. 
3. We wait until all clients and users of the existing implementation migrate to use the new API.
4. We deprecate existing implementation.

This still keeps my question open regarding what should we do with https://github.com/paritytech/json-rpc-interface-spec/pull/184, my preference here is that we make an exception and merge it, despite the fact that does not respect the naming convention, since it is an implementation that will be in use for awhile, alternatively it can live on my branch of this repo, but I consider that less ideal.

Let me know what you think!

josepot:
> * We have polkadot-sdk node and smoldot implement this version alongside the existing implementation.

Please, please, please stop using this as an argument. It really isn't.

> This still keeps my question open regarding what should we do with #184, my preference here is that we make an exception and merge it

Please, do **not** do that. There really is no reason to. Just like we never added into the spec any of the ad-hoc RPCs that were implemented without properly discussing their APIs. This really shouldn't be an exception. Also, it's not like we didn't try to convince Parity to create an RFC on the spec of those APIs and to have a proper discussion like we have done with all the modern JSON-RPC APIs.

The fact that someone decided to create an ad-hoc implementation/API can not be used as an argument to merge a set of APIs that we already know are less than ideal.

#184 should be closed and we should have a proper discussion about the APIs that we want in this PR. We are more than willing to make adjustments into what's proposed in here: changing names, reconsidering trade-offs, etc. However, please stop using as an argument an ad-hoc implementation... Just because that implementation exists, that doesn't make it a good API. Also, those APIs are not even compliant with this spec. Therefore, it makes no sense to "port" them here.It makes a lot more sense to iron out the APis that we want to have in the spec. Once the different clients implement the APIs of the spec, then we can adjust the tooling accordingly. Well, in fact, if we have a spec, we can actually create all the tooling and the SDKs even before they are implemented anywhere, that's actually one of the great advantages of having a spec.

If Parity doesn't have cycles to make these adjustments into the PolkadotSDK-node and into smoldot, we can 100% handle that.

alexggh:
> > * We have polkadot-sdk node and smoldot implement this version alongside the existing implementation.
> 
> Please, please, please stop using this as an argument. It really isn't.
> 
> > This still keeps my question open regarding what should we do with #184, my preference here is that we make an exception and merge it
> 
> Please, do **not** do that. There really is no reason to. Just like we never added into the spec any of the ad-hoc RPCs that were implemented without properly discussing their APIs. This really shouldn't be an exception. Also, it's not like we didn't try to convince Parity to create an RFC on the spec of those APIs and to have a proper discussion like we have done with all the modern JSON-RPC APIs.
> 
> The fact that someone decided to create an ad-hoc implementation/API can not be used as an argument to merge a set of APIs that we already know are less than ideal.
> 
> #184 should be closed and we should have a proper discussion about the APIs that we want in this PR. We are more than willing to make adjustments into what's proposed in here: changing names, reconsidering trade-offs, etc. 

I was proposing a middle ground here, since I thought it is better than the alternative, where we do not have them in this repo, although the implementation exists and they are used.

But I get your point, so I'm not going to insist anymore, I will  close https://github.com/paritytech/json-rpc-interface-spec/pull/184 and will focus on reviewing #185, and we can talk about implementation timelines once everyone is happy with this proposal.

src/api/statement_unstable_add.md
carlosala: sounds good, I have no particular love for names

src/api/statement.md
carlosala: The idea is that you're _following_ the store's state in your particular node. If you still prefer _subscribe_, it's ok for me, it makes sense _follow_ IMO but _subscribe_ sounds good.

src/api/statement_unstable_add.md
carlosala: The idea under `replyStatements` is that, every time you add a filter, the server will return the matching statements for that filter from the store. After this repetition (i.e. statements that are not new), statements will be sent just once, even if multiple filters match them.
- `remaining` at `replyStatements` might make sense. Nevertheless, I'd rather avoid it, not sure if it has any value.
- `remaining` at `newStatements` exposes an implementation detail. Some implementations might send statements right away, while others might buffer them and batch them. It has no value IMO.

src/api/statement_unstable_remove.md
carlosala: same as above

src/api/statement_unstable_unfollow.md
carlosala: same as above

src/api/statement_unstable_follow.md
carlosala: I advocate for not adding this. It complicates the whole situation: error management, implementation complexity, etc. I'd keep them separated.

carlosala:
Hey @alexggh, thanks for your review. I commented on your review items.

Regarding the legacy API, it is not a problem if it stays implemented. The idea about this one is to properly specify it and make it properly implementable in load balancers, light clients, etc. The same applies to the legacy "chainHead" API; it is still implemented and supported, but the modern one is preferred.

We would love to see more reviews and start implementing. As Josep mentioned, we can handle the implementation if Parity has no bandwidth.

src/api/statement_unstable_add.md
alexggh: > The idea under replyStatements is that, every time you add a filter, the server will return the matching statements for that filter from the store.

Alright, I understand the problem now, so because we will have to send the statements on the same subscription and we promise any "newStatements" will be sent once and only once, you need a way to separate the fact this statement is coming just as a reason of a new filter being added.

I would still relax the promise that any newStatements is sent only after in-progress replyStatements finishes, the arguments being:

1. There is noting special about statements being in the store and they should be treated the same as newStatements.
2. It makes the implementation more performant, because we can serve the subscriptions as new statements arrive, without having to store them somewhere and having to backfill them once replyStatements finishes.

src/api/statement_unstable_follow.md
alexggh: Could you develop more, I don't get where would all the complexity would come from since I don't imagine you want to start subscription before you have a filter. 

So on user side it shouldn't be a problem and then on node side would be just validating the filter and returning success or error with the code that you already have for add_filter.

The small advantage I see, is that you don't need two calls if you just a filter to subscribe, but it is small :).

src/api/statement.md
alexggh: If it is all the same to you let's call it subscribe :D. Thanks!

src/api/statement_unstable_add.md
carlosala: I don't think we should relax this requirement. The idea is that a filter is not active until `replayDone` is emitted, and it would be difficult to track otherwise. I still think it is preferable to maintain proper notification ordering.

src/api/statement_unstable_follow.md
carlosala: It is simply a separation of concerns. For example, there are specific errors for `follow` (now called `subscribe`) and for `add_filter`. Mixing them just for the sake of "avoiding one call" would be a bad practice, since in the future new functionality could be added to the `follow` subscription without touching filter / newStatements.

As a summary, having as simple APIs as possible is generally a good practice.

src/api/statement_unstable_add.md
alexggh: If we have a subscription with F1 and F2 active and a new filter F3 is added before we sent a replayDone, what behaviour are you proposing, if a new statement(S) is received that matches all filters.

1. Send newStatements containing S with F1 and F2 as filter Ids.
2. After we finished all replays for F3, send newStatements again with S, with F3 as filter id ?

carlosala: Nope, this would not match the spec.

> A newly inserted statement is emitted at most once per follow through `newStatements`, even if multiple active filters match it.

You'd wait until F3 is `done` and then emit it once for the 3 filters.

carlosala:
@alexggh done the renames 👍🏻

src/api/statement_unstable_follow.md
alexggh: I think we can both find plenty of subscription API's that allow you to both pass the filter in the initial call and to modify the  subscription afterwards, so I wouldn't call those "bad practice" :D. 

Anyways, the gain is small, so I'm fine with not passing a filter when creating the subscription.

src/api/statement_unstable_add.md
alexggh: I'll think about this one a bit more, I do think we pay some price in implementation complexity and don't see any gain, user of the API should not care about anything  other than "I will provide you some statements right away", there is no ordering in the statements to warrant us having to specifically mark them as "replyStatements"

carlosala: Marking them as `replyStatements` is only to break the guarantee of `newStatements` that a statement is emitted (at most) once for the duration of the subscription. It also helps load balancers to send the right notifications to the right consumers, making it easy for them to implement it with a variety of backends behind it.

alexggh: > It also helps load balancers to send the right notifications to the right consumers

Not sure I understand the load balancer argument, this would be the same subscription it would go through the load balancer to the same place for all added filters regardless if they are "replayStatements" or "newStatements", what am I missing ? 

> Marking them as replyStatements is only to break the guarantee of newStatements that a statement is emitted (at most) once for the duration of the subscription.

Thinking out loud, maybe that one we need to relax, since newStatements carry the filter-id we can have it as  "a statement is emitted(at most) once for any given filter id"

src/api/statement_unstable_subscribe.md
```diff
@@ -0,0 +1,154 @@
...
+
+For a given successful `statement_unstable_add_filter`, all statements that were already present in the store when the server accepted that call and that match its filter must be emitted before the corresponding `replayDone`.
+
+Statements that enter the store after a given successful `statement_unstable_add_filter` has been accepted must never be emitted through the `replayStatements` of that `statement_unstable_add_filter`, even if they match its filter.
+
+If a newly inserted statement matches one or more filters whose replays are still in progress on the same subscription, then this statement must not be emitted through `newStatements` until all these matching in-progress replays have emitted their corresponding `replayDone`, have been removed with `statement_unstable_remove_filter`, or have been interrupted by a `stop`.
```
DenzelPenzel, line 136: Should we mention in the spec either bound the wait (and emit stop if exceeded) or at least call out that overflow should result in stop?

carlosala: I don't think we should mention any of that.
If I understood you correctly, you propose to add some rules around stop. For instance, if I can't fulfill the notification in 3s, stop; if I have 1000 pending statements, stop.

This is, in my opinion, wrong. `stop` is not specified by design: every implementor might have their own limits, implementations, flaws and situations. `stop` means "I can't fulfill the spec right now", which might be due to a wide variety of problems that cannot be known at specification time.

Please clarify your point if I didn't get you right.

DenzelPenzel: You read me right, thanks!

src/api/statement_unstable_submit.md
```diff
@@ -0,0 +1,155 @@
...
+
+```json
+{
+  "status": "new"
+}
+```
+
+The statement was accepted and wasn't already present in the store.
+
+### known
```
AndreiEres, line 25: KnownExpired is missing

carlosala: `KnownExpired` would be incorrect for this API. It is an implementation detail of the polkadot-sdk store: polkadot-sdk keeps a list of statements that **have previously been** in the store but are no longer present because they were replaced by other statements.

For this API, the returned code should be `alreadyExpired` in that case. Whether the statement was ever present in the store is irrelevant to the consumer; the only relevant fact is that, at this point, it has expired and cannot be included.

AndreiEres: Forget it, clients never receive KnownExpired

AndreiEres:
Overall this looks like a good API for frontends. The shared-subscription design is especially nice — multiple client apps can multiplex filters onto one subscription, which means less load on the node.

Before merging though, I think we should do an implementation pass on our side to see what the real costs are. If some turn out to be high, I'd rather negotiate trade-offs on the spec — even ones that push a bit of work (e.g. dedup) to client apps — than ship a design that ends up unreliable in practice. A slightly less elegant API we can run reliably is the better outcome.

carlosala:
Sounds good, try to implement the API and let's review any findings. Feel free to contact me on Element if you have any trouble or question.

DenzelPenzel:
Ref to the API implementation https://github.com/paritytech/polkadot-sdk/pull/11989

DenzelPenzel:
> Ref to the API implementation [paritytech/polkadot-sdk#11989](https://github.com/paritytech/polkadot-sdk/pull/11989)

hi @carlosala we merged the API spec implementation.

FYI @josepot @alexggh

lexnv:
@DenzelPenzel is the implementation from https://github.com/paritytech/polkadot-sdk/pull/11989 fully aligned with this spec? Could you do a second review of this proposal so we get confirmation from your side before merging?

carlosala:
Thanks! I'll review the implementation and raise PRs for potential mistakes.

DenzelPenzel:
> @DenzelPenzel is the implementation from [paritytech/polkadot-sdk#11989](https://github.com/paritytech/polkadot-sdk/pull/11989) fully aligned with this spec? Could you do a second review of this proposal so we get confirmation from your side before merging?

@lexnv yes, the spec is good to merge as-is; one moment I see, that for `statement_unstable_submit`  `{"status": "known"}` is currently unreachable.  RPC submits with `StatementSource::Local`, which is allowed to resubmit, I will raise this topic with team, maybe we should fix the implementation to return `known` for local resubmission of a currently-present statement.

lexnv:
Oki doki, once we have that sorted out we can do a final pass over this PR 🙏

carlosala:
I added a small clarification, empty matchAll is not allowed.

DenzelPenzel (approved):

carlosala:
I found some flaws and opened a PR in https://github.com/paritytech/polkadot-sdk/pull/12706

src/api/statement_unstable_subscribe.md
```diff
@@ -0,0 +1,154 @@
...
+
+As a consequence, a statement that is emitted through `replayStatements` for a given `filterId` must not later be emitted through `newStatements` with that same `filterId`.
+
+If a subscription emits `stop` before replay completion, then no `replayDone` is required for the filters whose replay hasn't completed.
+
+JSON-RPC servers are allowed to delay live notifications in order to preserve the guarantees above.
+
+## Multiple subscriptions
+
+The JSON-RPC server must accept at least 2 `statement_unstable_subscribe` subscriptions per JSON-RPC client. Trying to open more might lead to a JSON-RPC error when calling `statement_unstable_subscribe`. In other words, as long as a JSON-RPC client starts 2 or fewer `statement_unstable_subscribe` subscriptions, it is guaranteed that this return value will never happen.
```
lexnv, line 148: This aligns with chainHead, however chainHead should support at least 2 because users are encouraged to create:
- one subscription `with_runtime = false` to receive updates faster
- one subscription `with_runtime = true` if they are interested in following the runtime 

Would update this to at least 1 if there's no other conflicting params (and the code on polkadot-sdk would remain spec complient even with 2 implementations)

carlosala: I have no particular problems with downsizing that to 1. I took it from CH directly.

If there are no comments against it I'll change it.

lexnv:
The spec looks solid! Just looking at it to see if there's any way to simplify the API surface a bit 🙏 


IIUC, when users start the subscription no events are emitted until filters are added via `statement_unstable_add_filter`. Because there's currently no way to pre-seed the filters, users are forced to make several `add_filter` calls immediately afterwards.

If we were to fix this by adding `seed_filter` to the subscription, we'd then have to figure out how to propagate the newly generated filter IDs back to the user. This makes me question if we really need the filter IDs at all 🤔 

Here's my mental model of the flow

### Current API

Users have to manage the add/remove filters:

```rust
// Subscribe to an empty subscription
let sub_id = statement_subscribe()

// Make several add_filter calls and track their IDs
let filter_id_1 = add_filter(sub_id, match_only {0x1, 0x2})
let filter_id_2 = add_filter(sub_id, match_only {0x1})


while let (notif) = sub_id.next().await {
    // Subscription propagates events with their associated filter IDs:
    // T0: { statement: 0x1, filter: {1, 2} }
    // T1: { statement: 0x2, filter: {1} }
}

// To update filters the user must reference the specific ID
remove_filter(filter_id_2)

// At this point, the subscription still produces `{ statement: 0x1, filter: {1} }`
```

### Alternative API

Could we let the user manage the filter state on their end and remove the concept of the fitler_id from the API?
- no concept of filter ID, users get back just statements
- We replace `add_filter` and `remove_filter` with a single `set_filter`
- we allow the preseeding with `statement_subscribe(filters: {})`

```rust
// no need for the prior add filter
// subscription could start with match empty (ie no events) if the `statements` field is not specified
let sub_id = statement_subscribe(statements: {0x1, 0x2})

// Later we only care about 0x1
set_filter(sub_id, statements: {0x1})
```

Another questions that I have around the usage:
- It is entirely possible that filters enter the store multiple times? (ie someother peer propagated the statement to us and we'll announce that through the subscription layer)
- Do statements live forever in our context? (ie is there an upper bound of subscriptions we keep around and new subscriptions would evict older ones, raising the need of a `dropStatement` event?)

@carlosala @DenzelPenzel would love to get your feedback on this since I dont have the full picture of the statements 🙏

carlosala:
Hey @lexnv, thanks for your comments.

1. Pre seeding filters: in my opinion it is a micro optimisation that does not pay off. It's simply separation of concerns, less error management and calls that do one thing, just to avoid a really small message. I wouldn't do so.
2. That filter_id situation is interesting. IMO it is nice to let the server inform about which filters match the statement sent, which is done by filter_id. It is computation that the server needs to do in order to fulfill the spec and the client avoids it by using the information sent back by the server. I'm proposing an API that works as "operations", that you can cancel and start whenever you want, and you're proposing an API that is a single "operation", that essentially you cancel and restart (by `set_filter`) every time you want to change any part of it. I don't see then how would we manage replay statements and stuff, I don't see the advantage of it honestly.
3. Filters might be repeated, yes, there is no limitation to it by spec. Servers though might set a limitation to it and emit stop if that happens, that is entirely let to implementation.
4. `drop_statement` is not needed. Statements are ephemeral, they are not meant to store data that I evict whenever I see fit. The notification to the consumer is "this statement is in my store at this point", not "this statement stays in my store and I'll tell you when it leaves". The rules for storing statements are not enforced, there is no single "store" coherent across nodes; servers might just see statements pass and never save them on disk (light-clients, for instance) and therefore adding that `drop_statement` requirement with certain rules would limit highly low-resource clients.

lexnv:
@carlosala Thanks for the context! The spec looks good as is! 

One other thing before merging I'd like to double check with you:

Could we replace the concept of the `filter_id` entirely with the `subscription_id` (which we get from jsonrpsee management already)? 
The API would look like:
 - `subscribe_statement([statements of interest)]` + `unsubscribe` (no other methods on this subscription)
 
 The previous example:
 
 ```rust
let filter_id_1 = add_filter(sub_id, match_only {0x1, 0x2})
let filter_id_2 = add_filter(sub_id, match_only {0x1})
```

Would effectively turn into two subscription (short / medium lived):

```rust
let sub_id1 = subscribe({0x1, 0x2})
let sub_id2 = subscribe({0x1})
```

This also has the benefit of simplifying the events received. Then we'd bump the number of subscriptions allowed per client to the number of filter_ids we support currently 🤔

DenzelPenzel:
> @carlosala Thanks for the context! The spec looks good as is!
> 
> One other thing before merging I'd like to double check with you:
> 
> Could we replace the concept of the `filter_id` entirely with the `subscription_id` (which we get from jsonrpsee management already)? The API would look like:
> 
> * `subscribe_statement([statements of interest)]` + `unsubscribe` (no other methods on this subscription)
> 
> The previous example:
> 
> ```rust
> let filter_id_1 = add_filter(sub_id, match_only {0x1, 0x2})
> let filter_id_2 = add_filter(sub_id, match_only {0x1})
> ```
> 
> Would effectively turn into two subscription (short / medium lived):
> 
> ```rust
> let sub_id1 = subscribe({0x1, 0x2})
> let sub_id2 = subscribe({0x1})
> ```
> 
> This also has the benefit of simplifying the events received. Then we'd bump the number of subscriptions allowed per client to the number of filter_ids we support currently 🤔



> @carlosala Thanks for the context! The spec looks good as is!
> 
> One other thing before merging I'd like to double check with you:
> 
> Could we replace the concept of the `filter_id` entirely with the `subscription_id` (which we get from jsonrpsee management already)? The API would look like:
> 
> * `subscribe_statement([statements of interest)]` + `unsubscribe` (no other methods on this subscription)
> 
> The previous example:
> 
> ```rust
> let filter_id_1 = add_filter(sub_id, match_only {0x1, 0x2})
> let filter_id_2 = add_filter(sub_id, match_only {0x1})
> ```
> 
> Would effectively turn into two subscription (short / medium lived):
> 
> ```rust
> let sub_id1 = subscribe({0x1, 0x2})
> let sub_id2 = subscribe({0x1})
> ```
> 
> This also has the benefit of simplifying the events received. Then we'd bump the number of subscriptions allowed per client to the number of filter_ids we support currently 🤔

@lexnv worth naming what this converges to: `subscribe([topics]) + unsubscribe` with no filter
ids is the shape of our existing v1 API (`statement_subscribeStatement`). So the contested question
reduces to one thing: was multi-filter multiplexing worth it at all?

And `filter_id` buys exactly one thing a client can't reconstruct: a statement matching K filters is
delivered once instead of K times

lexnv (approved):
LGTM! Thanks @carlosala for adding this spec! 🙏

PR #186, opened 2026-05-07
bitswap_unstable: stream added

michalkucharczyk:
This PR adds: [`bitswap_unstable_stream`](https://github.com/paritytech/json-rpc-interface-spec/blob/mku-bitswap-stream/src/api/bitswap_unstable_stream.md)

...and renames `bitswap_v1_get` to [`bitswap_unstable_get`](https://github.com/paritytech/json-rpc-interface-spec/blob/mku-bitswap-stream/src/api/bitswap_unstable_get.md)


src/api/bitswap_v1_getMany.md
```diff
@@ -0,0 +1,83 @@
+# bitswap_v1_getMany
+
+**Parameters**:
+
+- `cids`: Array of CID strings, in [string format](https://github.com/multiformats/cid/blob/edb1c5294ad2d8257812d7ded4941c3e0fafccf3/README.md#variant---stringified-form).  
+  Each entry must satisfy the same constraints as the `cid` parameter of [`bitswap_v1_get`](bitswap_v1_get.md): CIDv1 in `base32` multibase encoding (string starting from `b...`), with the `sha2-256` or `blake2b-256` hash function.  
+  The maximum number of CIDs accepted in a single request is implementation-defined; an implementation must accept at least 16 CIDs and may accept more.
```
bkontur: > The maximum number of CIDs accepted in a single request is implementation-defined;

What does happen when the input "maximum number of CIDs" is exceeded? Also `-32602 InvalidParams` or dedicated error code?

src/api/bitswap_v1_getMany.md
```diff
@@ -0,0 +1,83 @@
...
+```json
+{ "code": -32810, "message": "..." }
+```
+
+`code` carries the same [error categories](bitswap_v1_get.md#error-categories) as the top-level error
+of [`bitswap_v1_get`](bitswap_v1_get.md). Clients that already know how to interpret `code` for
+`bitswap_v1_get` can reuse the same retry logic per-CID.
+
+`message` is a human-readable diagnostic string for logs and developer-facing tooling. Only `code` is
+stable for programmatic dispatch.
```
bkontur:
```suggestion
stable for programmatic dispatch.
```

src/api/bitswap_v1_getMany.md
bkontur: ok, I see PoC and there is `TooManyCids`, so we should align/adjust https://github.com/paritytech/json-rpc-interface-spec/blob/main/src/api/bitswap_v1_get.md#error-categories

michalkucharczyk: `TooManyCids` will be in the error message, as other input error params. We don't have per-inner-error mapping to RPC errors. It was already done for `get`. See: https://github.com/paritytech/polkadot-sdk/blob/bcd139f350ac253b4cc0409993e1d070c4bcc396/substrate/client/rpc-spec-v2/src/bitswap/error.rs#L64-L90

I think it is fine, but it is not strong opinion. With higher _granularity_ we will need to keep RPC specification updated. Umbrella error is probably more flexible.

michalkucharczyk: Second thought: yeah maybe it is worth aligning. Will add it.

src/api/bitswap.md
```diff
@@ -9,3 +9,19 @@ nodes.
 
 The full node may also implement downloading the data from other full nodes via Bitswap protocol
 if it doesn't have the data locally (e.g., due to the block pruning settings).
+
+The namespace exposes four methods:
+
+- [`bitswap_v1_get`](bitswap_v1_get.md) — fetch a single chunk by CID. Errors are reported at the
```
lexnv: Initially, we expose the RPC methods as `bitswap_unstable_*`.However, if we are confident this wont change its fine to go directly with a V1 🙏 

For the chainhead we adopted the `unstable` for about 6months to give PJS/Papi time to provide some feedback before the final stabilization

src/api/bitswap_v1_stream.md
```diff
@@ -0,0 +1,93 @@
...
+[ "<cid>", { "code": -32810, "message": "..." } ]
+```
+
+`code` carries the same [error categories](bitswap_v1_get.md#error-categories) as the top-level error of [`bitswap_v1_get`](bitswap_v1_get.md). Clients that already know how to interpret `code` for `bitswap_v1_get` can reuse the same retry logic per-CID.
+
+`message` is a human-readable diagnostic string for logs and developer-facing tooling. Only `code` is stable for programmatic dispatch.
+
+A missing or invalid CID in the input produces an `Err` event for that CID and does not abort the rest of the stream. Because events are emitted in arrival order, an `Err` event for one CID may be emitted before, after, or interleaved with `Ok` events for other CIDs.
+
+## Example notifications
```
lexnv: We could provide a bit more info for subscriptions to align better with chainHead / archive.

Events for the subscriptions:

- Retrieval of CIDs on the happy path:

```json
{
    "event": "queryResult",
    "cid": ..,
    "value": "0x...",
}
```

- Error while retrieving the CID


```json
{
    "event": "queryError",
    "cid": ..,
    "error_code":.., 
    "message": ...,
}
```

- Event emitted at the end:

```json
{
    "event": "queryDone",
}
```

src/api/bitswap.md
```diff
@@ -9,3 +9,19 @@ nodes.
...
+  JSON-RPC level.
+- [`bitswap_v1_getMany`](bitswap_v1_getMany.md) — fetch a batch of chunks in a single call. The
+  response is an array with one outcome per input CID, in input order. Per-CID errors are embedded
+  in the response so that one failure does not abort the rest of the batch.
+- [`bitswap_v1_stream`](bitswap_v1_stream.md) — like `bitswap_v1_getMany`, but exposes the per-CID
```
lexnv: I would suggest removing the `bitswap_v1_getMany` / `bitswap_v1_get` methods entirely, since the same results can be obtained with `bitswap_v1_stream`, which provides better backpressure and reduces memory usage when processing queries.

Then the client side could create a thin wrapper:

```rust
fn bitswap_v1_getMany {
  let events = bitswap_v1_stream.subscribe();
  while Some(event) = events.next().await {
      result.push(event);
  }
 return result;
}
```

For example, this would align beter with archive_storage: https://github.com/paritytech/json-rpc-interface-spec/blob/main/src/api/archive_v1_storage.md. We are exposing a subscription that allows querying multiple items. There are no `storage_get` / `storage_many` method alternatives since those methods could be wrapped on the client side easily 

src/api/bitswap_v1_stream.md
michalkucharczyk: Do we need `queryDone`? The count of events is known, so techincally client knows when the stream is terminated?

michalkucharczyk: Also the current form is in this very shape because event is aligned with the result of `getMany` - the same handler can be used to parse.

I don't have opinion which is better.

src/api/bitswap.md
michalkucharczyk: The internal spec were this was discussed did not conclude on removal of `get/getMany`.  I'd agree that stream alone makes more sense.

src/api/bitswap.md
michalkucharczyk: I was just following v1_get pattern :)

src/api/bitswap.md
skunert: I would keep `get`, because its specified and implemented. Agree on `bitswap_v1_getMany`, it does not seem useful.

src/api/bitswap_v1_getMany.md
```diff
@@ -0,0 +1,70 @@
...
+
+The response array preserves input order. The caller can re-issue the call with just `<cidB>`.
+
+## Possible errors
+
+See [error categories](bitswap_v1_get.md#error-categories).
+
+## Empty input
+
+An empty input array (`cids: []`) is allowed and returns an empty result array. No CID is fetched.
```
skunert: Why?

src/api/bitswap_v1_stream.md
```diff
@@ -0,0 +1,93 @@
...
+
+The client may then call `bitswap_v1_unstream` to release the subscription, or simply stop reading.
+
+## Possible errors
+
+Whole-call failures cause the subscription to be rejected (no events are emitted). See [error categories](bitswap_v1_get.md#error-categories).
+
+## Empty input
+
+An empty input array (`cids: []`) is allowed. The subscription opens, no events are emitted, and the client can release it via `bitswap_v1_unstream` whenever convenient.
```
skunert: Same here, why do we allow this?

src/api/bitswap_v1_getMany.md
```diff
@@ -0,0 +1,70 @@
...
+
+Same encoding as the return value of [`bitswap_v1_get`](bitswap_v1_get.md).
+
+**Err**:
+
+```json
+{ "code": -32810, "message": "..." }
+```
+
+`code` carries the same [error categories](bitswap_v1_get.md#error-categories) as the top-level error of [`bitswap_v1_get`](bitswap_v1_get.md). Clients that already know how to interpret `code` for `bitswap_v1_get` can reuse the same retry logic per-CID.
```
skunert: What do you think about any timing guarantees? Like when I provide 3 CIDs and 2 arrive, then I wait for one more event. As consumer might be nice to know that an event is guaranteed to arrive within 20 seconds or something. Either a result or an error.

Happy to hear opinions, not sure if we need this.

src/api/bitswap.md
skunert: Hmmm yeah maybe it was not a good choice for the v1_get too. At least there the API is super simple. unstable seems like the more cautious choice to me.

src/api/bitswap_v1_getMany.md
michalkucharczyk: No reason. We can forbid it. Both ways are fine for me.

src/api/bitswap_v1_getMany.md
michalkucharczyk: If you want specific timeout you can implement it on your side. We should not make any assumptions about network topology/speed in API?

src/api/bitswap.md
michalkucharczyk: @skunert @lexnv I'll rename `v1_stream` to `unstable_stream`.
And I will drop `getMany`.

~I am not going to touch `v1_get` - as it is already used and deployed here and there?~

src/api/bitswap_v1_getMany.md
michalkucharczyk: done, see: https://github.com/paritytech/json-rpc-interface-spec/blob/mku-bitswap-stream/src/api/bitswap_unstable_stream.md#possible-errors

src/api/bitswap.md
michalkucharczyk: Done.

src/api/bitswap_v1_stream.md
michalkucharczyk: Done. See: https://github.com/paritytech/json-rpc-interface-spec/blob/mku-bitswap-stream/src/api/bitswap_unstable_stream.md#notifications-format

src/api/bitswap.md
michalkucharczyk: Done, `getMany` removed.

src/api/bitswap_v1_getMany.md
michalkucharczyk: Done, you got a section here: https://github.com/paritytech/json-rpc-interface-spec/blob/mku-bitswap-stream/src/api/bitswap_unstable_stream.md#empty-input :)

src/api/bitswap_v1_stream.md
michalkucharczyk: Same: https://github.com/paritytech/json-rpc-interface-spec/blob/mku-bitswap-stream/src/api/bitswap_unstable_stream.md#empty-input

lexnv (dismissed):
LGTM!

src/api/bitswap.md
```diff
@@ -1,11 +1,11 @@
 # Introduction
 
-The functions with `bitswap` prefix allow fetching data chunks from the storage given their CID,
-normally downloaded using Bitswap protocol.
+The functions with `bitswap` prefix allow fetching data chunks  given their CID, normally downloaded using Bitswap protocol.
```
dmitry-markin:
```suggestion
The functions with `bitswap` prefix allow fetching data chunks given their CID, normally downloaded using Bitswap protocol.
```

src/api/bitswap_unstable_get.md
```diff
@@ -39,23 +39,18 @@ Error category `FailRetryBackoff` can be retried after a delay. For example, suc
...
+logs and developer-facing tooling, is implementation-specific, and must not be relied upon in
+business logic.
 
-### Detailed error information
+### Example JSON-RPC success response
 
-| Field          | Description                          | Example value                         |
-| -------------- | ------------------------------------ | ------------------------------------- |
-| `message`      | Human-readable error description     | `Request timeout.`                    |
-| `data.variant` | Error variant for structured logging | `Timeout`, `NoPeers`, `NotFound`, ... |
```
dmitry-markin, line 51: Why have you decided to purge `variant`?

dmitry-markin (approved):

src/api/bitswap_unstable_get.md
dmitry-markin:
```suggestion
  Only `sha2-256`, `blake2b-256` & `keccak-256` hash functions are supported.  
```
Let's also include `keccak-256` to not create a separate PR, it is supported by smoldot and a polkadot-sdk PR is out: https://github.com/paritytech/polkadot-sdk/pull/12163.

dmitry-markin (dismissed):

dmitry-markin:
Actually, I don't think we should remove `bitswap_v1_get`. Polkadot app is likely already using it.

Introducing `bitswap_unstable_get` also can be postponed until we have to break `bitswap_v1_get`.

src/api/bitswap_unstable_get.md
michalkucharczyk: Two main reasons:
- it provided redundant information,
- it was first method using this pattern in this repository - so I decided to align with other methods,

michalkucharczyk:
> Actually, I don't think we should remove `bitswap_v1_get`. Polkadot app is likely already using it.
> 
> Introducing `bitswap_unstable_get` also can be postponed until we have to break `bitswap_v1_get`.

Alias will be kept - so `bitswap_v1_get` method will be working. 
I also discussed it with @bkontur - app shall be updated, we still should have time to do it.

lexnv:
> Alias will be kept - so bitswap_v1_get method will be working. 
 I also discussed it with @bkontur - app shall be updated, we still should have time to do it.

Perfect, ideally once we are sure the RPC method is stable (ie, no further modifications like changing / removing / adding hashig types and battle tested to gain feedback), we can turn this into a stable_v1 🙏

lexnv (approved):
LGTM! Thanks for tackling the event formats for the steam implementation 🙏

dmitry-markin (approved):

